PR #1, opened 2021-12-11
Parse semicolon

helixbass:
In this PR:
- parse file consisting of a single semicolon

If you want to compare against the Typescript compiler source code, I've got `v4.5.2` checked out

To test:
If you have a file named eg `tmp.tsx` in the repo root directory whose contents are just a semicolon (no trailing newline - if your editor wants to save it with a trailing newline, you can then run `truncate -s -1 tmp.tsx` to remove the final byte from the file) and run `cargo run tmp.tsx` you should see debug-logging of a single `SourceFile` whose `statements` contains a single `EmptyStatement`


src/compiler/parser.rs
```diff
@@ -0,0 +1,306 @@
...
+        self.next_token_without_check()
+    }
+
+    fn is_identifier(&self) -> bool {
+        false
+    }
+
+    fn parse_expected(&mut self, kind: SyntaxKind, should_advance: Option<bool>) -> bool {
+        let should_advance = should_advance.unwrap_or(true);
+        if is_same_variant(&self.token(), &kind) {
```
helixbass, line 174: I figured this might not work but I initially tried writing this as `matches!(self.token(), kind)` (so comparing two "dynamic" enum variants), and interestingly it didn't flag it as a compiler error, I just noticed that something was wrong because it was warning that `kind` was unused. So I guess the `matches!` macro was expanding it to something nonsensical but syntactically valid

So then per [this answer](https://stackoverflow.com/a/32554326/732366), added an `is_same_variant()` helper for this kind of dynamic comparison of enum variants (discriminant-only)

src/compiler/parser.rs
```diff
@@ -0,0 +1,306 @@
...
+
+        self.parse_error_at_current_token(&Diagnostics::_0_expected);
+        false
+    }
+
+    fn create_node_array(&self, elements: Vec<Node>) -> NodeArray {
+        self.factory.create_node_array(elements)
+    }
+
+    fn finish_node<TParsedNode: NodeInterface>(&mut self, node: TParsedNode) -> TParsedNode {
```
helixbass, line 189: So `Node` is now a concrete enum type, but sort of per above comment re shared "base node" fields, there are places where we want to treat whichever concrete node subtype "as a node" (eg to manipulate its shared "base node" fields). So I exposed `NodeInterface` to represent that (so eg `EmptyStatement` implements `NodeInterface`), which would consist of getters + setters for those base node fields

src/compiler/program.rs
```diff
@@ -0,0 +1,169 @@
...
+    file_name: &str,
+) -> Option<Rc<SourceFile>> {
+    let _path = to_path(helper_context, file_name);
+
+    let file = helper_context.host.get_source_file(file_name);
+
+    file.map(|file| {
+        let file = Rc::new(file);
+        helper_context.processing_other_files.push(file.clone());
+        file
```
helixbass, line 152: The reason that I made these `SourceFile`'s wrapped in `Rc`'s is that here `findSourceFileWorker()` both appends a newly-created `SourceFile` to `processing_other_files` and returns it. So it seems like I couldn't make `processing_other_files` a `Vec` of _references_ (`Vec<&SourceFile>`) because a reference from here would only live until `find_source_file_worker()` returns, nor a `Vec<Box<SourceFile>>` because a box takes ownership of the boxed value (so then I couldn't return it from `find_source_file_worker()` too), and it didn't seem like a good idea to make `SourceFile`'s `Clone`-able (and so store a separate cloned copy in `processing_other_files`) since they contain entire AST's. So I used `Rc`'s

src/compiler/scanner.rs
```diff
@@ -0,0 +1,141 @@
...
+        self.set_text_(text);
+        self.set_end(match length {
+            None => text.len(),
+            Some(length) => start.unwrap() + length,
+        });
+        self.set_text_pos(start.unwrap_or(0));
+    }
+
+    pub fn set_text_pos(&mut self, text_pos: usize) {
+        // Debug_.assert(text_pos >= 0);
```
helixbass, line 64: Rust helpfully pointed out that this assertion would never fail with `text_pos` typed as `usize`

src/compiler/scanner.rs
```diff
@@ -0,0 +1,141 @@
...
+    // fn scan_identifier(&self, start_character: char) -> {
+    // }
+}
+
+pub fn create_scanner() -> Scanner {
+    Scanner::new()
+}
+
+fn code_point_at(s: &str, i: usize) -> char {
+    s.chars().nth(i).unwrap()
```
helixbass, line 140: This is egregiously inefficient, in order to scan the file character-by-character I'm iterating through all previous characters until I get to the one I want. If it actually makes sense to always treat the source text as just _characters_ (vs _bytes_), then maybe converting `String` -> `Vec[char]` and then indexing into that `Vec` would make sense?

src/compiler/sys.rs
```diff
@@ -0,0 +1,45 @@
...
+
+    fn read_file(&self, file_name: &str) -> Option<String> {
+        self.read_file_worker(file_name)
+    }
+
+    fn get_current_directory(&self) -> String {
+        env::current_dir()
+            .unwrap()
+            .into_os_string()
+            .into_string()
```
helixbass, line 32: Since I'm using `String` everywhere right now, apparently this is how to convert from a `PathBuf`/`Path` to a `String`

src/compiler/types.rs
```diff
@@ -0,0 +1,185 @@
...
+
+pub struct CreateProgramOptions<'config> {
+    pub root_names: &'config [String],
+}
+
+#[non_exhaustive]
+pub struct CharacterCodes;
+#[allow(non_upper_case_globals)]
+impl CharacterCodes {
+    pub const asterisk: char = '*';
```
helixbass, line 142: In Typescript these are numbers (as hex codes), per above comment this is related to what we're treating the fundamental units of source text as I think (chars/bytes/Unicode code points)

src/lib.rs
```diff
@@ -0,0 +1,28 @@
+#[macro_use]
+extern crate lazy_static;
+
+mod compiler;
+mod execute_command_line;
+mod rust_helpers;
+
+pub use compiler::command_line_parser::parse_command_line;
```
helixbass, line 8: In the Typescript code, all of the helpers etc from different files live under a "flat" `namespace ts` so having them be exposed as top-level names under the crate seems like the equivalent?

src/main.rs
```diff
@@ -1,3 +1,6 @@
+use tsc_rust::{execute_command_line, get_sys, System};
+
 fn main() {
```
helixbass, line 3: This corresponds to the body of `src/tsc/tsc.ts` in the Typescript code

src/compiler/core.rs
```diff
@@ -0,0 +1,16 @@
...
+    array: &[TItem],
+    callback: &mut dyn FnMut(&TItem, usize) -> Option<TReturn>,
+) -> Option<TReturn> {
+    for (index, item) in array.iter().enumerate() {
+        let result = callback(item, index);
+        if let Some(value) = result {
+            return Some(value);
+        }
+    }
+    None
```
peterstuart, line 11: I think this loop is equivalent to this:

```rs
array
    .iter()
    .enumerate()
    .find_map(|(index, item)| callback(item, index))
```

src/compiler/debug.rs
```diff
@@ -0,0 +1,24 @@
...
+        panic!("{}", message);
+    }
+
+    pub fn assert(&self, expression: bool) {
+        if !expression {
+            let message = "False expression.";
+            self.fail(Some(message));
+        }
+    }
+}
```
peterstuart, line 20: Since the `DebugType` struct doesn't really do anything, I'd probably make these free functions that you call using the module name to qualify them:

```rust
debug::assert(false)
```

src/compiler/diagnostic_information_map_generated.rs
```diff
@@ -0,0 +1,27 @@
...
+pub struct Diagnostics;
+#[allow(non_upper_case_globals)]
+impl Diagnostics {
+    pub const _0_expected: DiagnosticMessage = diag(
+        1005,
+        DiagnosticCategory::Error,
+        "_0_expected_1005",
+        "'{0}' expected.",
+    );
+}
```
peterstuart, line 27: Similar to `Debug`, I'm not sure why this is a struct rather than just functions.

src/compiler/factory/node_factory.rs
```diff
@@ -0,0 +1,50 @@
+use crate::{
+    BaseNode, BaseNodeFactory, EmptyStatement, NodeArray, NodeArrayOrVec, NodeFactory, SourceFile,
+    SyntaxKind,
+};
+
+impl NodeFactory {
+    pub fn create_node_array<TElements: Into<NodeArrayOrVec>>(
```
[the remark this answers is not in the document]
peterstuart: Rather than having the intermediate `Into<NodeArrayOrVec>`, could you have `TElements: Into<NodeArray>`? `NodeArray` could implement `From<NodeArray>` (if that's not automatic) and `From<Vec<Node>>`.

src/compiler/factory/node_factory.rs
```diff
@@ -0,0 +1,50 @@
...
+            NodeArrayOrVec::Vec(elements) => NodeArray::new(elements),
+        }
+    }
+
+    pub fn create_empty_statement<TBaseNodeFactory: BaseNodeFactory>(
+        &self,
+        base_factory: &TBaseNodeFactory,
+    ) -> EmptyStatement {
+        EmptyStatement {
+            _node: self.create_base_node(base_factory, SyntaxKind::EmptyStatement),
```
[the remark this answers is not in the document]
peterstuart: If I'm understanding correctly, I think there is option (c) which I'd go for if it works:

```rust
// in node.rs

enum Type {
  Statement(Statement)
}

struct Node {
  base: BaseNode
  type: Type
}

// in statement.rs

enum Statement {
  Empty
}
```

src/compiler/parser.rs
```diff
@@ -0,0 +1,306 @@
...
+    SourceFileConstructor: Option<fn(SyntaxKind) -> BaseNode>,
+    factory: NodeFactory,
+    file_name: Option<String>,
+    parse_diagnostics: Option<Vec<DiagnosticWithDetachedLocation>>,
+    current_token: Option<SyntaxKind>,
+    parse_error_before_next_finished_node: bool,
+}
+
+impl ParserType {
+    fn new() -> ParserType {
```
peterstuart, line 29: I'd usually do `-> Self` instead of the type name there.

src/compiler/parser.rs
```diff
@@ -0,0 +1,306 @@
...
+
+        self.scanner.set_text(Some(_source_text), None, None);
+    }
+
+    fn parse_source_file_worker(&mut self) -> SourceFile {
+        self.next_token();
+
+        let statements =
+            self.parse_list(ParsingContext::SourceElements, ParserType::parse_statement);
+        Debug_.assert(matches!(self.token(), SyntaxKind::EndOfFileToken));
```
peterstuart, line 107: You could use the build in `assert!` macro here instead.

src/compiler/parser.rs
```diff
@@ -0,0 +1,306 @@
...
+        &mut self,
+        start: usize,
+        length: usize,
+        message: &DiagnosticMessage,
+    ) {
+        let last_error = last_or_undefined(self.parse_diagnostics());
+        if match last_error {
+            None => true,
+            Some(last_error) => last_error.start != start,
+        } {
```
peterstuart, line 140: You can use the `map_or` function to make this a bit more concise:

```rs
if last_error.map_or(true, |last_error| last_error.start != start) {
  ...
}
```

src/compiler/parser.rs
```diff
@@ -0,0 +1,306 @@
...
+            self.parse_error_before_next_finished_node = false;
+        }
+
+        node
+    }
+
+    fn is_list_element(&self, kind: ParsingContext) -> bool {
+        match kind {
+            ParsingContext::SourceElements => self.is_start_of_statement(),
+            _ => panic!("Unimplemented"),
```
peterstuart, line 200: There is actually an `unimplemented` macro which is a variant of `panic` you can use:

```rust
unimplemented!()
```

src/compiler/parser.rs
```diff
@@ -0,0 +1,306 @@
...
+        }
+    }
+
+    fn is_list_terminator(&self) -> bool {
+        if matches!(self.token(), SyntaxKind::EndOfFileToken) {
+            return true;
+        }
+        false
+    }
```
peterstuart, line 209: I'd derive `Eq, PartialEq` for `SyntaxKind` so you can write this function like this:

```rs
fn is_list_terminator(&self) -> bool {
    return self.token() == SyntaxKind::EndOfFileToken;
}
```

src/compiler/parser.rs
```diff
@@ -0,0 +1,306 @@
...
+        &mut self,
+        _parsing_context: ParsingContext,
+        parse_element: fn(&mut ParserType) -> TItem,
+    ) -> TItem {
+        parse_element(self)
+    }
+
+    fn is_start_of_left_hand_side_expression(&self) -> bool {
+        match self.token() {
+            _ => self.is_identifier(),
```
peterstuart, line 239: What is the `match` for here?

src/compiler/path.rs
```diff
@@ -0,0 +1,129 @@
+use crate::{last_or_undefined, CharacterCodes, Path};
```
peterstuart, line 1: This file could mostly be replaced with https://doc.rust-lang.org/stable/std/path/index.html (which is presumably pretty robustly tested and cross-platform).

src/compiler/program.rs
```diff
@@ -0,0 +1,169 @@
...
+
+    let mut processing_other_files: Option<Vec<Rc<SourceFile>>> = None;
+    let mut files: Vec<Rc<SourceFile>> = vec![];
+
+    let host = create_compiler_host();
+
+    let current_directory = host.get_current_directory();
+
+    let structure_is_reused = StructureIsReused::Not;
+    if !matches!(structure_is_reused, StructureIsReused::Completely) {
```
peterstuart, line 96: I'd use `==` here too rather than `matches!`.

src/compiler/scanner.rs
```diff
@@ -0,0 +1,141 @@
...
+        new_text: Option<&str>,
+        start: Option<usize>,
+        length: Option<usize>,
+    ) {
+        let text = new_text.unwrap_or("");
+        self.set_text_(text);
+        self.set_end(match length {
+            None => text.len(),
+            Some(length) => start.unwrap() + length,
+        });
```
peterstuart, line 59: I'd use `map_or` here too

src/compiler/scanner.rs
peterstuart: That seems like an improvement, but I think ideally you would iterate over `s.chars()` directly, and let it keep track of the indexing for you.

src/compiler/sys.rs
```diff
@@ -0,0 +1,45 @@
...
+
+lazy_static! {
+    static ref SYS: SystemConcrete = SystemConcrete {
+        args: env::args().skip(1).collect(),
+    };
+}
+
+pub fn get_sys() -> &'static impl System {
+    &*SYS
+}
```
peterstuart, line 45: What does this get you over directly exposing `SYS` from the `lazy_static`?

src/compiler/types.rs
```diff
@@ -0,0 +1,185 @@
...
+#[derive(Debug)]
+pub struct NodeArray {
+    _nodes: Vec<Node>,
+}
+
+impl NodeArray {
+    pub fn new(nodes: Vec<Node>) -> Self {
+        NodeArray { _nodes: nodes }
+    }
+}
```
peterstuart, line 52: If the only point of this is to be consistent with the TS naming, you could use a type alias to `Vec<Node>` rather than wrapping it in a struct, which would get rid of all the converting stuff.

src/compiler/types.rs
```diff
@@ -0,0 +1,185 @@
...
+
+impl From<EmptyStatement> for Statement {
+    fn from(empty_statement: EmptyStatement) -> Self {
+        Statement::EmptyStatement(empty_statement)
+    }
+}
+
+#[derive(Debug)]
+pub struct SourceFile {
+    pub _node: BaseNode,
```
peterstuart, line 109: I haven't looked at the TS code, but I'd have assumed that the `_` in front of `node` meant it should be a private field?

src/compiler/types.rs
```diff
@@ -0,0 +1,185 @@
...
+}
+
+#[non_exhaustive]
+pub struct CharacterCodes;
+#[allow(non_upper_case_globals)]
+impl CharacterCodes {
+    pub const asterisk: char = '*';
+    pub const semicolon: char = ';';
+    pub const slash: char = '/';
+}
```
peterstuart, line 145: Same question as above somewhere: these seem like they could be public `const`s in a module used like `character_code::asterisk` rather than in a struct.

src/lib.rs
peterstuart: Yep, and I think that's a pretty typical pattern in rust either way.

peterstuart (commented):
I didn't go deep on the equivalent TS code, but added a bunch of Rust-specific comments.

src/compiler/factory/node_factory.rs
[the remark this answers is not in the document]
peterstuart: Actually, just put the fields from `BaseNode` directly on `Node`.

src/compiler/path.rs
helixbass: Ya this one could be interesting depending on especially how differences between `String`/`OsString`/`Path` are relevant for real use cases

src/compiler/program.rs
helixbass: Ok updated to derive `Eq, PartialEq` for `StructureIsReused` and use `==` here

src/compiler/scanner.rs
helixbass: Ok updated to use `map_or`

src/compiler/scanner.rs
helixbass: I don't think the scanner literally only does a single pass of the file (eg there's lookahead) so I don't think it could just gradually consume a single iterator

src/compiler/sys.rs
helixbass: I ran into the fact that I guess `lazy_static!` actually [creates a one-off type under the hood](https://stackoverflow.com/a/48115258/732366) for the static variable which then meant that I couldn't type `SYS` as implementing `System` (because the one-off type doesn't "inherit" the trait implementations of the declared type I guess)

src/compiler/types.rs
helixbass: There will be other fields added to `NodeArray`

src/compiler/types.rs
helixbass: This doesn't mimic the Typescript structure so I used `_node` to sort of denote that

src/compiler/factory/node_factory.rs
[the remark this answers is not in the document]
helixbass: > If I'm understanding correctly, I think there is option (c) which I'd go for if it works

Let me play around with this one separately (I'll go ahead and merge this PR, I'd make those changes against the current branch to minimize merge conflicts)

PR #2, opened 2021-12-12
Parse number

helixbass:
In this PR:
- parse numeric literal

To test:
If you have a file named eg `tmp.tsx` in the repo root directory whose contents are just a number eg `12` (no trailing newline) and run `cargo run tmp.tsx` you should see debug-logging of a single `SourceFile` whose `statements` contains a single `ExpressionStatement` whose `expression` is a `NumericLiteral` with `text: "12"`

### DIFF
--- a/src/compiler/factory/node_factory.rs
+++ b/src/compiler/factory/node_factory.rs
@@ -1,6 +1,6 @@
 use crate::{
-    BaseNode, BaseNodeFactory, EmptyStatement, Identifier, NodeArray, NodeArrayOrVec, NodeFactory,
-    SourceFile, SyntaxKind,
+    BaseNode, BaseNodeFactory, EmptyStatement, Expression, ExpressionStatement, Identifier,
+    NodeArray, NodeArrayOrVec, NodeFactory, SourceFile, SyntaxKind,
 };
 
 impl NodeFactory {
@@ -42,6 +42,17 @@ impl NodeFactory {
     ) -> EmptyStatement {
         EmptyStatement {
             _node: self.create_base_node(base_factory, SyntaxKind::EmptyStatement),
+        }
+    }
+
+    pub fn create_expression_statement<TBaseNodeFactory: BaseNodeFactory>(
+        &self,
+        base_factory: &TBaseNodeFactory,
+        expression: Expression,
+    ) -> ExpressionStatement {
+        ExpressionStatement {
+            _node: self.create_base_node(base_factory, SyntaxKind::ExpressionStatement),
+            expression,
         }
     }
 
--- a/src/compiler/factory/node_factory.rs
+++ b/src/compiler/factory/node_factory.rs
@@ -1,6 +1,6 @@
 use crate::{
-    BaseNode, BaseNodeFactory, EmptyStatement, NodeArray, NodeArrayOrVec, NodeFactory, SourceFile,
-    SyntaxKind,
+    BaseNode, BaseNodeFactory, EmptyStatement, Identifier, NodeArray, NodeArrayOrVec, NodeFactory,
+    SourceFile, SyntaxKind,
 };
 
 impl NodeFactory {
@@ -12,6 +12,28 @@ impl NodeFactory {
             NodeArrayOrVec::NodeArray(node_array) => node_array,
             NodeArrayOrVec::Vec(elements) => NodeArray::new(elements),
         }
+    }
+
+    fn create_base_identifier<TBaseNodeFactory: BaseNodeFactory>(
+        &self,
+        base_factory: &TBaseNodeFactory,
+        text: &str,
+    ) -> Identifier {
+        let node = base_factory.create_base_identifier_node(SyntaxKind::Identifier);
+        let node = Identifier {
+            _node: node,
+            escaped_text: text.to_string(),
+        };
+        node
+    }
+
+    pub fn create_identifier<TBaseNodeFactory: BaseNodeFactory>(
+        &self,
+        base_factory: &TBaseNodeFactory,
+        text: &str,
+    ) -> Identifier {
+        let node = self.create_base_identifier(base_factory, text);
+        node
     }
 
     pub fn create_empty_statement<TBaseNodeFactory: BaseNodeFactory>(
--- a/src/compiler/factory/node_factory.rs
+++ b/src/compiler/factory/node_factory.rs
@@ -1,6 +1,7 @@
 use crate::{
-    BaseNode, BaseNodeFactory, EmptyStatement, Expression, ExpressionStatement, Identifier,
-    NodeArray, NodeArrayOrVec, NodeFactory, SourceFile, SyntaxKind,
+    BaseLiteralLikeNode, BaseNode, BaseNodeFactory, EmptyStatement, Expression,
+    ExpressionStatement, Identifier, NodeArray, NodeArrayOrVec, NodeFactory, NumericLiteral,
+    SourceFile, SyntaxKind,
 };
 
 impl NodeFactory {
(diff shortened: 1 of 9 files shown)


src/compiler/parser.rs
```diff
@@ -194,6 +234,37 @@ impl ParserType {
         node
     }
 
+    fn create_missing_node(
```
helixbass, line 237: The `createMissingNode()` helper returns multiple different AST node subtypes depending on what it's passed so I represented that with a wrapping `MissingNode` enum that I then unwrap below to extract the inner `Identifier`

src/compiler/parser.rs
```diff
@@ -194,6 +234,37 @@ impl ParserType {
...
+        &mut self,
+        kind: SyntaxKind,
+        diagnostic_message: DiagnosticMessage,
+    ) -> MissingNode {
+        self.parse_error_at_current_token(&diagnostic_message);
+
+        let result = MissingNode::Identifier(self.factory.create_identifier(self, ""));
+        self.finish_node(result)
```
helixbass, line 245: ...and since I implemented `NodeInterface` for `MissingNode` it can be passed to `finish_node()`

src/compiler/scanner.rs
```diff
@@ -33,6 +43,24 @@ impl Scanner {
...
+                | CharacterCodes::_4
+                | CharacterCodes::_5
+                | CharacterCodes::_6
+                | CharacterCodes::_7
+                | CharacterCodes::_8
+                | CharacterCodes::_9 => {
+                    let ScanNumberReturn {
+                        type_: token,
+                        value: token_value,
+                    } = self.scan_number();
```
helixbass, line 59: The more natural way to represent these one-off compound return types in Rust would presumably be to return a tuple but I'm trying to mimic the Typescript code as closely as I can, so creating these one-off struct types for structured return values

src/compiler/types.rs
```diff
@@ -8,13 +8,16 @@ impl Path {
     }
 }
 
-#[derive(Copy, Clone, Debug)]
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
```
helixbass, line 11: I saw that you were recommending deriving `Eq`/`PartialEq` to be able to use `==` to compare enum variants

src/compiler/utilities.rs
```diff
@@ -1,19 +1,71 @@
...
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub enum OperatorPrecedence {
+    Comma,
+    Multiplicative,
+    Primary,
+    Invalid = -1,
+}
+
+impl OperatorPrecedence {
+    pub const Lowest: OperatorPrecedence = OperatorPrecedence::Comma;
```
helixbass, line 16: This seems like a nice way of translating the pattern in the Typescript code where they have "marker" enum values that refer to one of the other enum values (so here I can say `OperatorPrecedence::Lowest` which is of type `OperatorPrecedence` even though it's not actually one of the `OperatorPrecedence` enum variants)

src/compiler/utilities.rs
```diff
@@ -1,19 +1,71 @@
...
+}
+
+impl Ord for OperatorPrecedence {
```
helixbass, line 19: In `parseBinaryExpressionRest()` it compares `OperatorPrecedence`'s by their numeric enum value so I implemented `Ord`/`PartialOrd` to be able to write it similarly in Rust (eg `new_precedence > precedence`)

src/compiler/scanner.rs
```diff
@@ -128,6 +157,63 @@ impl Scanner {
...
+        ch >= CharacterCodes::_0 && ch <= CharacterCodes::_9
+    }
+
+    fn scan_number_fragment(&mut self) -> String {
+        let start = self.pos();
+        let result = "".to_string();
+        loop {
+            let ch = self.text().chars().nth(self.pos());
+            let ch = match ch {
+                Some(ch) => ch,
```
peterstuart, line 178: You could move the following if-statement into the match as a guard, which I think would be more clear. (Doing this off the top of my head)

```rust
match ch {
  Some(ch) if self.is_digit(ch) = > {
    self.set_pos(self.pos() + 1);
    continue;
  },
  None => break;
}

File: src/compiler/scanner.rs [hunk 8]
```diff
@@ -128,6 +157,63 @@ impl Scanner {
...
+                .collect::<String>(),
+        );
+        ret
+    }
+
+    fn scan_number(&mut self) -> ScanNumberReturn {
+        let start = self.pos();
+        let main_fragment = self.scan_number_fragment();
+        let end = self.pos();
+        let result: String = self.text().chars().skip(start).take(end - start).collect();
```
peterstuart, line 203: I haven't done it, but it seems like you should be able to use the range operator for this rather than having to skip + take.

src/compiler/utilities.rs
peterstuart: I think you could derive rather than implement this if (although you might need to rearrange it so `Invalid` was first)?

src/compiler/parser.rs
```diff
@@ -246,17 +335,116 @@ impl ParserType {
...
+                break;
+            }
         }
+
+        left_operand
     }
 
-    // fn parse_expression(&self) -> Box<dyn Expression> {}
+    fn is_binary_operator(&self) -> bool {
+        get_binary_operator_precedence(self.token()) > OperatorPrecedence::Comma
```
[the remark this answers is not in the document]
peterstuart: It'd be much more clear to implement this on `OperatorPrecendence` itself, I think.

src/compiler/factory/node_factory.rs
```diff
@@ -14,6 +15,60 @@ impl NodeFactory {
         }
     }
 
+    fn create_base_literal<TBaseNodeFactory: BaseNodeFactory>(
+        &self,
+        base_factory: &TBaseNodeFactory,
+        kind: SyntaxKind,
+        value: &str,
+    ) -> BaseLiteralLikeNode {
```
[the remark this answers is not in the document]
helixbass: Ya per Slack the thing that would seem to lose is the ability for a helper to return eg `BinaryExpression` as a concrete type that includes both the base node fields and the additional subtype-specific fields

src/compiler/parser.rs
```diff
@@ -246,17 +335,116 @@ impl ParserType {
...
+
+                self.is_identifier()
+            }
+        }
+    }
+
+    fn parse_expression(&mut self) -> Expression {
+        let expr = self.parse_assignment_expression_or_higher();
+
+        expr
```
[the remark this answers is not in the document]
helixbass: Ya just mimicking Typescript code where there will be intermediate logic

src/compiler/parser.rs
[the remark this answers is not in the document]
helixbass: Ya in the Typescript code it's actually written as `> 0` since I guess you can mix/match number-valued enums + actual numbers, but so here I'm explicitly referencing the equivalent enum variant (which I think is actually clearer anyway). Ya basically `getBinaryOperatorPrecedence()` either returns a precedence that's higher than `OperatorPrecedence.Comma` or returns `-1` (ie `OperatorPrecedence.Invalid`) so does seem equivalent to just saying "is the current token in the set of binary operators?"

src/compiler/scanner.rs
helixbass: Ya this `match` doesn't exist in the Typescript version because if we reach the end of the source text (while parsing a number), `.charCodeAt()` returns `NaN` which can be passed along to the other helpers which are expecting a `number`, whereas in Rust they're expecting `char`'s rather than `Option<char>`'s so I need to "unwrap" (and stop parsing the number if it's `None`). I actually like these patterns where you can unwrap eg an `Option` without having to nest the unwrapped logic by having the `None` branch abort the control flow (eg by `panic!()`'ing or here `break`'ing)

src/compiler/scanner.rs
helixbass: This gets into the text-as-chars vs text-as-bytes thing, indexing a string with a range gives you a range of bytes rather than chars I believe

src/compiler/utilities.rs
helixbass: Right that would be my understanding from reading the [docs](https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html#derivable). In this case I'd like to mimic the Typescript ordering with `Invalid` last

PR #3, opened 2021-12-12
Parse number multiplication

helixbass:
In this PR:
- parse multiplication of integer literals

To test:
If you have a file named eg `tmp.tsx` in the repo root directory whose contents are just a multiplication of two integer literals eg `12 * 3` (no trailing newline) and run `cargo run tmp.tsx` you should see debug-logging of a single `SourceFile` whose `statements` contains a single `ExpressionStatement` whose `expression` is a `BinaryExpression` with `left` as `NumericLiteral` with `text: "12"`, `operator_token` as `Token` with `kind: AsteriskToken`, and `right` as `NumericLiteral` with `text: "3"`

Based on `parse-asterisk`

### DIFF
--- a/src/compiler/factory/node_factory.rs
+++ b/src/compiler/factory/node_factory.rs
@@ -1,7 +1,7 @@
 use crate::{
-    BaseLiteralLikeNode, BaseNode, BaseNodeFactory, EmptyStatement, Expression,
-    ExpressionStatement, Identifier, NodeArray, NodeArrayOrVec, NodeFactory, NumericLiteral,
-    SourceFile, SyntaxKind,
+    BaseLiteralLikeNode, BaseNode, BaseNodeFactory, BinaryExpression, EmptyStatement, Expression,
+    ExpressionStatement, Identifier, Node, NodeArray, NodeArrayOrVec, NodeFactory, NumericLiteral,
+    SourceFile, SyntaxKind, Token,
 };
 
 impl NodeFactory {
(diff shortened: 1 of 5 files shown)


src/compiler/factory/node_factory.rs
```diff
@@ -69,6 +69,41 @@ impl NodeFactory {
         base_factory.create_base_token_node(kind)
     }
 
+    pub fn create_token<TBaseNodeFactory: BaseNodeFactory>(
+        &self,
+        base_factory: &TBaseNodeFactory,
+        token: SyntaxKind,
+    ) -> Token {
```
helixbass, line 76: Even though a `Token` doesn't have any fields beyond a `Node`/`BaseNode`, again (like `Expression`) it felt right for `Token` to be a top-level variant of `Node` (where `Token`'s are things like punctuation and keywords). But then I made `Token` a "leaf" type (as opposed to `Expression`, which is an enum type) because I don't think any of its subtypes add any fields

src/compiler/types.rs
```diff
@@ -112,6 +134,26 @@ impl From<Expression> for Node {
     }
 }
 
+#[derive(Debug)]
+pub struct BinaryExpression {
+    pub _node: BaseNode,
+    pub left: Box<Expression>,
```
helixbass, line 140: Add `Box` indirection to avoid circularity

src/compiler/factory/node_factory.rs
```diff
@@ -69,6 +69,41 @@ impl NodeFactory {
...
+        operator: Node,
+        right: Expression,
+    ) -> BinaryExpression {
+        let node = self.create_base_expression(base_factory, SyntaxKind::BinaryExpression);
+        let node = BinaryExpression {
+            _node: node,
+            left: Box::new(left),
+            operator_token: Box::new(operator),
+            right: Box::new(right),
+        };
```
peterstuart, line 103: For something like this where the boxing is an implementation detail of `BinaryExpression`, I'd probably hide that behind a constructor so you can do `BinaryExpression::new(node, left, operator, right)`.

src/compiler/scanner.rs
```diff
@@ -6,6 +6,7 @@ struct ScanNumberReturn {
 }
 
 pub struct Scanner {
+    skip_trivia: bool,
```
peterstuart, line 9: What's this do?

src/compiler/types.rs
peterstuart: Is this a situation where it'd have an unbounded size without this?

src/compiler/scanner.rs
helixbass: Seems like it controls whether the scanner throws away whitespace vs somehow paying closer attention to it. In this instance of the scanner owned by the parser it looks like it's always set to `true` so not sure when it's in the other mode

src/compiler/types.rs
helixbass: Ya since `Expression` has an enum variant that contains a `BinaryExpression` I think trying to have `BinaryExpression` contain an `Expression` would make the size unbounded

PR #4, opened 2021-12-13
Parse boolean multiplication

helixbass:
In this PR:
- support parsing a multiplication of integer literals and/or boolean literals

To test:
If you have a file named eg `tmp.tsx` in the repo root directory whose contents are just a multiplication of an integer literal and a boolean literal eg `12 * true` (no trailing newline) and run `cargo run tmp.tsx` you should see debug-logging of a single `SourceFile` whose `statements` contains a single `ExpressionStatement` whose `expression` is a `BinaryExpression` with `left` as `NumericLiteral` with `text: "12"`, `operator_token` as `BaseNode` with `kind: AsteriskToken`, and `right` as `TokenExpression` with `kind: TrueKeyword`

### DIFF
--- a/src/compiler/factory/node_factory.rs
+++ b/src/compiler/factory/node_factory.rs
@@ -1,7 +1,7 @@
 use crate::{
     BaseLiteralLikeNode, BaseNode, BaseNodeFactory, BinaryExpression, EmptyStatement, Expression,
     ExpressionStatement, Identifier, Node, NodeArray, NodeArrayOrVec, NodeFactory, NumericLiteral,
-    SourceFile, SyntaxKind, Token,
+    SourceFile, SyntaxKind,
 };
 
 impl NodeFactory {
@@ -73,9 +73,9 @@ impl NodeFactory {
         &self,
         base_factory: &TBaseNodeFactory,
         token: SyntaxKind,
-    ) -> Token {
+    ) -> BaseNode {
         let node = self.create_base_token(base_factory, token);
-        Token { _node: node }
+        node
     }
 
     fn create_base_expression<TBaseNodeFactory: BaseNodeFactory>(
--- a/src/compiler/parser.rs
+++ b/src/compiler/parser.rs
@@ -219,11 +219,10 @@ impl ParserType {
         false
     }
 
-    fn parse_token_node(&mut self) -> Node {
+    fn parse_token_node(&mut self) -> BaseNode {
         let kind = self.token();
         self.next_token();
         self.finish_node(self.factory.create_token(self, kind))
-            .into()
     }
 
     fn create_node_array(&self, elements: Vec<Node>) -> NodeArray {
@@ -387,7 +386,7 @@ impl ParserType {
             let operator_token = self.parse_token_node();
             let right = self.parse_binary_expression_or_higher(new_precedence);
             left_operand = self
-                .make_binary_expression(left_operand, operator_token, right)
+                .make_binary_expression(left_operand, operator_token.into(), right)
                 .into();
         }
 
@@ -455,6 +454,9 @@ impl ParserType {
     fn parse_primary_expression(&mut self) -> Expression {
         match self.token() {
             SyntaxKind::NumericLiteral => return self.parse_literal_node().into(),
+            SyntaxKind::TrueKeyword | SyntaxKind::FalseKeyword => {
+                return self.parse_token_node().into()
+            }
             _ => (),
         }
 
--- a/src/compiler/parser.rs
+++ b/src/compiler/parser.rs
@@ -386,7 +386,7 @@ impl ParserType {
             let operator_token = self.parse_token_node();
             let right = self.parse_binary_expression_or_higher(new_precedence);
             left_operand = self
-                .make_binary_expression(left_operand, operator_token.into(), right)
+                .make_binary_expression(left_operand, operator_token, right)
                 .into();
         }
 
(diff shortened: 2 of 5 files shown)


src/compiler/factory/node_factory.rs
```diff
@@ -73,9 +73,9 @@ impl NodeFactory {
         &self,
         base_factory: &TBaseNodeFactory,
         token: SyntaxKind,
-    ) -> Token {
+    ) -> BaseNode {
```
helixbass, line 76: I ran into an issue with my `Token`/`Expression` as enum variants of `Node` hierarchy - Typescript blurs the line a bit eg calling `createTokenNode<PrimaryExpression>()`. So I preserved `Expression` but removed `Token` in favor of just `BaseNode` as a variant of `Node` (and to represent "basic" concrete expression types like boolean literals, I added a `TokenExpression` variant to `Expression`). So that way these things that were returning `Token` can return `BaseNode`, which can then be hoisted into an `Expression::TokenExpression()` (when we're constructing expressions like boolean literals)

src/compiler/parser.rs
```diff
@@ -387,7 +386,7 @@ impl ParserType {
             let operator_token = self.parse_token_node();
             let right = self.parse_binary_expression_or_higher(new_precedence);
             left_operand = self
-                .make_binary_expression(left_operand, operator_token, right)
+                .make_binary_expression(left_operand, operator_token.into(), right)
```
helixbass, line 389: Here this is calling `BaseNode::into::<Node>()` (which wraps in a `Node::BaseNode`)...

src/compiler/parser.rs
```diff
@@ -455,6 +454,9 @@ impl ParserType {
     fn parse_primary_expression(&mut self) -> Expression {
         match self.token() {
             SyntaxKind::NumericLiteral => return self.parse_literal_node().into(),
+            SyntaxKind::TrueKeyword | SyntaxKind::FalseKeyword => {
+                return self.parse_token_node().into()
```
helixbass, line 458: ...and here this is calling `BaseNode::into::<Expression>()` (which wraps in an `Expression::TokenExpression`)

src/compiler/scanner.rs
```diff
@@ -1,5 +1,44 @@
...
+lazy_static! {
+    static ref text_to_keyword_obj: HashMap<&'static str, SyntaxKind> =
+        HashMap::from_iter(IntoIter::new([
+            ("false", SyntaxKind::FalseKeyword),
+            ("true", SyntaxKind::TrueKeyword),
+        ]));
+}
+
+lazy_static! {
+    static ref text_to_keyword: HashMap<&'static str, SyntaxKind> = text_to_keyword_obj.clone();
```
helixbass, line 18: In the Typescript version `textToKeywordObj` is a JS object while `textToKeyword` is an ES6 `Map`, so here they're just both `HashMap`'s

src/compiler/scanner.rs
```diff
@@ -1,5 +1,44 @@
...
+}
+
+fn is_unicode_identifier_start(ch: char) -> bool {
+    false
+}
+
+fn is_identifier_start(ch: char) -> bool {
+    ch >= CharacterCodes::A && ch <= CharacterCodes::Z
+        || ch >= CharacterCodes::a && ch <= CharacterCodes::z
+        || ch == CharacterCodes::dollar_sign
```
helixbass, line 28: In something a bit meta, in the Typescript version this is called `CharacterCodes::$`/`CharacterCodes::_` because those are allowed as "identifier starts" in TS/JS, whereas in Rust I guess they're not (or solitary `_` is reserved/special?) so had to spell them out as `dollar_sign`/`underscore`

src/compiler/types.rs
```diff
@@ -306,7 +311,13 @@ pub struct CreateProgramOptions<'config> {
 pub struct CharacterCodes;
 #[allow(non_upper_case_globals)]
 impl CharacterCodes {
+    pub const max_ascii_character: char = '';
```
helixbass, line 314: Since these are specified as numbers in the Typescript version, I believe that I correctly typed in the equivalent (to `0x7f`) character here - maybe it would be more obvious what's going on to use some integer-to-char conversion helper here?

src/compiler/parser.rs
peterstuart: If you wanted to get fancy, you could make `make_binary_expression` have a type parameter of something convertible into a `Node`, and then you wouldn't need `into` at the call sites.

src/compiler/scanner.rs
```diff
@@ -1,5 +1,44 @@
...
+
+use std::array::IntoIter;
+use std::collections::HashMap;
+use std::iter::FromIterator;
+
 use crate::{CharacterCodes, SyntaxKind};
 
+lazy_static! {
+    static ref text_to_keyword_obj: HashMap<&'static str, SyntaxKind> =
+        HashMap::from_iter(IntoIter::new([
```
peterstuart, line 11: I think that if you put a `&` in front of this array, it'd be iterable, so you could use the more normal `into_iter` or whatever here.

src/compiler/scanner.rs
peterstuart: Ah yeah, `_` is the "ignore this" variable/param name in Rust, so that makes sense you can't use it.

src/compiler/scanner.rs
```diff
@@ -73,13 +126,43 @@ impl Scanner {
...
+                }
+                self.set_pos(self.pos() + char_size(ch));
+            }
+            self.set_token_value(
+                &self
+                    .text()
+                    .chars()
+                    .skip(self.token_pos())
+                    .take(self.pos() - self.token_pos())
+                    .collect::<String>(),
```
peterstuart, line 159: Is the `<String>` here required or for clarity? (I'm surprised it couldn't infer it, since it's being used directly as a param to `set_token_value`.)

src/compiler/parser.rs
helixbass: Sure, updated to accept a `<TNode: Into<Node>>`

src/compiler/scanner.rs
helixbass: I'm not sure exactly what that would look like, this was mimicking some answer re how to approximate a `HashMap` literal in Rust

src/compiler/scanner.rs
helixbass: If I omit the `<String>` it fails with ``value of type `str` cannot be built from `std::iter::Iterator<Item=char>` `` - I think the inference issue would be that `.set_token_value()` takes an `&str`, so without some "`String` deref's into `str`" "extra intelligence" it would want to assume that this expression that's being `&`'ed should be typed as `str`, not `String`

PR #5, opened 2021-12-17
Emit type error

helixbass:
In this PR:
- emit a type error when parsing + type-checking the prefix unary expression `++true`

To test:
If you have a file named eg `tmp.tsx` in the repo root directory whose contents are just `++true` (no trailing newline) and run `cargo run tmp.tsx` you should see debug-logging of the `SourceFile` and then a single `DiagnosticWithLocation` associated with the `SourceFile` whose `file_name` is `"tmp.tsx"` and with location data `start: 2, length: 4` (ie the location of the `true` operand)

### DIFF
--- a/src/compiler/binder.rs
+++ b/src/compiler/binder.rs
@@ -1,0 +1,57 @@
+#![allow(non_upper_case_globals)]
+
+use std::rc::Rc;
+
+use crate::{Node, SourceFile};
+
+// lazy_static! {
+//     static ref binder: BinderType = create_binder();
+// }
+
+pub fn bind_source_file(file: Rc<SourceFile>) {
+    // binder.call(file);
+    create_binder().call(file);
+}
+
+struct BinderType {
+    file: Option<Rc<SourceFile>>,
+    parent: Option<Rc<Node>>,
+}
+
+fn create_binder() -> BinderType {
+    BinderType {
+        file: None,
+        parent: None,
+    }
+}
+
+impl BinderType {
+    fn call(&self, f: Rc<SourceFile>) {
+        self.bind_source_file(f);
+    }
+
+    fn file(&self) -> Rc<SourceFile> {
+        self.file.unwrap().clone()
+    }
+
+    fn bind_source_file(&self, f: Rc<SourceFile>) {
+        self.file = Some(f);
+
+        if true {
+            self.bind(self.file());
+        }
+
+        self.file = None;
+        self.parent = None;
+    }
+
+    fn bind(&self, node: Option<Rc<Node>>) {
+        let node = match node {
+            None => {
+                return;
+            }
+            Some(node) => node,
+        };
+        set_parent(node);
+    }
+}
--- a/src/compiler/binder.rs
+++ b/src/compiler/binder.rs
@@ -2,7 +2,7 @@
 
 use std::rc::Rc;
 
-use crate::{Node, SourceFile};
+use crate::{set_parent, Node, SourceFile};
 
 // lazy_static! {
 //     static ref binder: BinderType = create_binder();
@@ -26,19 +26,19 @@ fn create_binder() -> BinderType {
 }
 
 impl BinderType {
-    fn call(&self, f: Rc<SourceFile>) {
+    fn call(&mut self, f: Rc<SourceFile>) {
         self.bind_source_file(f);
     }
 
     fn file(&self) -> Rc<SourceFile> {
-        self.file.unwrap().clone()
+        self.file.as_ref().unwrap().clone()
     }
 
-    fn bind_source_file(&self, f: Rc<SourceFile>) {
+    fn bind_source_file(&mut self, f: Rc<SourceFile>) {
         self.file = Some(f);
 
         if true {
-            self.bind(self.file());
+            self.bind(Some(Rc::new(self.file().into())));
         }
 
         self.file = None;
@@ -52,6 +52,12 @@ impl BinderType {
             }
             Some(node) => node,
         };
-        set_parent(node);
+        set_parent(
+            &*node,
+            match &self.parent {
+                None => None,
+                Some(parent) => Some(parent.clone()),
+            },
+        );
     }
 }
--- a/src/compiler/binder.rs
+++ b/src/compiler/binder.rs
@@ -6,7 +6,7 @@ use std::sync::RwLock;
 
 use crate::{
     for_each, for_each_child, set_parent, Expression, ExpressionStatement, Node, NodeArray,
-    NodeInterface, SourceFile, Statement, SyntaxKind,
+    NodeInterface, Statement, SyntaxKind,
 };
 
 bitflags! {
@@ -25,13 +25,13 @@ bitflags! {
 //     static ref binder: BinderType = create_binder();
 // }
 
-pub fn bind_source_file(file: Rc<SourceFile>) {
+pub fn bind_source_file(file: Rc<Node>) {
     // binder.call(file);
     create_binder().call(file);
 }
 
 struct BinderType {
-    file: RwLock<Option<Rc<SourceFile>>>,
+    file: RwLock<Option<Rc</*SourceFile*/ Node>>>,
     parent: RwLock<Option<Rc<Node>>>,
 }
 
@@ -43,15 +43,15 @@ fn create_binder() -> BinderType {
 }
 
 impl BinderType {
-    fn call(&self, f: Rc<SourceFile>) {
+    fn call(&self, f: Rc<Node>) {
         self.bind_source_file(f);
     }
 
-    fn file(&self) -> Rc<SourceFile> {
+    fn file(&self) -> Rc<Node> {
         self.file.read().unwrap().as_ref().unwrap().clone()
     }
 
-    fn set_file(&self, file: Option<Rc<SourceFile>>) {
+    fn set_file(&self, file: Option<Rc<Node>>) {
         *self.file.write().unwrap() = file;
     }
 
@@ -63,11 +63,11 @@ impl BinderType {
         *self.parent.write().unwrap() = parent;
     }
 
-    fn bind_source_file(&self, f: Rc<SourceFile>) {
+    fn bind_source_file(&self, f: Rc<Node>) {
         self.set_file(Some(f.clone()));
 
         if true {
-            self.bind(Some(Rc::new(self.file().into())));
+            self.bind(Some(self.file()));
         }
 
         self.set_file(None);
--- a/src/compiler/binder.rs
+++ b/src/compiler/binder.rs
@@ -1,8 +1,25 @@
 #![allow(non_upper_case_globals)]
 
+use bitflags::bitflags;
 use std::rc::Rc;
+use std::sync::RwLock;
 
-use crate::{set_parent, Node, SourceFile};
+use crate::{
+    for_each, for_each_child, set_parent, Expression, ExpressionStatement, Node, NodeArray,
+    NodeInterface, SourceFile, Statement, SyntaxKind,
+};
+
+bitflags! {
+    struct ContainerFlags: u32 {
+        const None = 0;
+
+        const IsContainer = 1 << 0;
+
+        const IsControlFlowContainer = 1 << 2;
+
+        const HasLocals = 1 << 5;
+    }
+}
 
 // lazy_static! {
 //     static ref binder: BinderType = create_binder();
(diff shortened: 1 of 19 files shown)


src/compiler/checker.rs
```diff
@@ -0,0 +1,363 @@
...
+        }
+    }
+
+    fn get_union_type(&self, types: Vec<Rc<Type>>) -> Rc<Type> {
+        let mut type_set: Vec<Rc<Type>> = vec![];
+        self.add_types_to_union(&mut type_set, /*TypeFlags::empty(), */ &types);
+        self.get_union_type_from_sorted_list(type_set)
+    }
+
+    fn get_union_type_from_sorted_list(&self, types: Vec<Rc<Type>>) -> Rc<Type> {
```
helixbass, line 173: Maybe I should make eg `SortedList` a concrete type so that I can enforce at the type level that I'm passing an already-sorted list since it seems to be distinguishing those (though I haven't implemented any of the actual sorting logic yet or corresponding eg insertion/lookup optimizations)

src/compiler/checker.rs
```diff
@@ -0,0 +1,363 @@
...
+        false
+    }
+
+    fn check_type_related_to(
+        &self,
+        source: Rc<Type>,
+        target: Rc<Type>,
+        relation: &HashMap<String, RelationComparisonResult>,
+    ) -> bool {
+        false
```
helixbass, line 244: This is what's actually triggering the type error on the `true` operand in `++true`. So I'd have to implement real "is `source` assignable to `target`?" logic here in order for eg `++1` to not get flagged but `++true` to get flagged

src/compiler/checker.rs
```diff
@@ -0,0 +1,363 @@
...
+
+    fn check_prefix_unary_expression(&self, node: &PrefixUnaryExpression) -> Rc<Type> {
+        let operand_expression = match &*node.operand {
+            Node::Expression(expression) => expression,
+            _ => panic!("Expected Expression"),
+        };
+        let operand_type = self.check_expression(operand_expression);
+        match node.operator {
+            SyntaxKind::PlusPlusToken => {
+                self.check_arithmetic_operand_type(operand_expression, operand_type.clone(), &Diagnostics::An_arithmetic_operand_must_be_of_type_any_number_bigint_or_an_enum_type);
```
helixbass, line 315: Further up the stack, this is where it's actually specifying which type error (aka "diagnostic") to emit if it thinks the argument to `++` isn't number-ish

src/compiler/checker.rs
```diff
@@ -0,0 +1,363 @@
...
+            Expression::TokenExpression(token_expression) => match token_expression.kind() {
+                SyntaxKind::TrueKeyword => {
+                    return self.true_type().clone();
+                }
+                _ => unimplemented!(),
+            },
+            Expression::PrefixUnaryExpression(prefix_unary_expression) => {
+                return self.check_prefix_unary_expression(prefix_unary_expression);
+            }
+            // Expression::BinaryExpression(binary_expression) => {
```
helixbass, line 343: As mentioned, didn't implement binary-expression type-checking first because there's a "trampolining" optimization happening under the hood

src/compiler/checker.rs
```diff
@@ -0,0 +1,363 @@
...
+        let expression = match &*node.expression {
+            Node::Expression(expression) => expression,
+            _ => panic!("Expected Expression"),
+        };
+        self.check_expression(expression);
+    }
+
+    fn initialize_type_checker<TTypeCheckerHost: TypeCheckerHost>(&self, host: &TTypeCheckerHost) {
+        for file in host.get_source_files() {
+            bind_source_file(file);
```
helixbass, line 360: This is the entry point to binding (so the type checker invokes the binder for each source file)

src/compiler/core.rs
```diff
@@ -1,13 +1,42 @@
-pub fn for_each<TItem, TReturn>(
-    array: &[TItem],
-    callback: &mut dyn FnMut(&TItem, usize) -> Option<TReturn>,
+use crate::SortedArray;
+
+pub fn for_each<
+    TCollection: IntoIterator,
```
helixbass, line 4: I think I changed this so that I can pass a `NodeArray` to `for_each()`? But I don't remember

src/compiler/core.rs
```diff
@@ -1,13 +1,42 @@
...
+    mut callback: TCallback,
 ) -> Option<TReturn> {
     array
-        .iter()
+        .into_iter()
         .enumerate()
         .find_map(|(index, item)| callback(item, index))
 }
 
+fn some<TItem>(array: &[TItem], predicate: Option<Box<dyn FnMut(&TItem) -> bool>>) -> bool {
```
helixbass, line 17: I think I should change things like this to be generic over the closure type rather than boxing the closure type - although here then if the caller is passing `None` for `predicate` the compiler would probably complain about not being able to resolve the closure generic?

src/compiler/core.rs
```diff
@@ -1,13 +1,42 @@
...
+        return array1;
+    }
+    if !some(&array1, None) {
+        return array2;
+    }
+    array1.append(&mut array2);
+    array1
+}
+
+pub fn insert_sorted<TItem>(array: &mut SortedArray<TItem>, insert: TItem) {
```
helixbass, line 32: Oh wait looks like I already do have a `SortedArray` type (which means the Typescript version does, so I wonder why the one I commented on above isn't taking a `SortedArray` based on its name?)

src/compiler/parser.rs
```diff
@@ -1,16 +1,40 @@
...
+    Diagnostics, Expression, Identifier, LiteralLikeNode, Node, NodeArray, NodeArrayOrVec,
+    NodeFactory, NodeInterface, OperatorPrecedence, ReadonlyTextRange, Scanner, SourceFile,
     Statement, SyntaxKind,
 };
 
+fn visit_node<TNodeCallback: FnMut(Rc<Node>)>(mut cb_node: TNodeCallback, node: Rc<Node>) {
+    cb_node(node)
+}
+
+pub fn for_each_child<TNodeCallback: FnMut(Rc<Node>), TNodesCallback: FnMut(&NodeArray)>(
```
helixbass, line 19: This appears to be sort of a generic AST node visitor helper

src/compiler/parser.rs
```diff
@@ -25,15 +49,53 @@ impl NodeInterface for MissingNode {
...
 }
 
 #[allow(non_snake_case)]
 struct ParserType {
     scanner: Scanner,
-    NodeConstructor: Option<fn(SyntaxKind) -> BaseNode>,
-    IdentifierConstructor: Option<fn(SyntaxKind) -> BaseNode>,
-    TokenConstructor: Option<fn(SyntaxKind) -> BaseNode>,
-    SourceFileConstructor: Option<fn(SyntaxKind) -> BaseNode>,
+    NodeConstructor: Option<fn(SyntaxKind, usize, usize) -> BaseNode>,
```
helixbass, line 95: Wiring up AST node location data (`pos`/`end` aka `ReadonlyTextRange`)

src/compiler/program.rs
```diff
@@ -40,53 +41,136 @@ fn create_compiler_host_worker() -> impl CompilerHost {
...
-        ProgramConcrete { files }
+    pub fn new(files: Vec<Rc<Node>>) -> Self {
+        ProgramConcrete {
+            files,
+            diagnostics_producing_type_checker: None,
+        }
+    }
+
+    fn get_diagnostics_producing_type_checker(&mut self) -> &TypeChecker {
+        // self.diagnostics_producing_type_checker
```
helixbass, line 57: This was two different ways that I tried to write this logic but where the borrow checker was unhappy (the pattern here is basically "return the already-cached value or initialize the cache and then return that newly created value")

src/compiler/program.rs
```diff
@@ -101,27 +185,31 @@ pub fn create_program(root_names_or_options: CreateProgramOptions) -> impl Progr
...
+
 fn process_root_file(helper_context: &mut CreateProgramHelperContext, file_name: &str) {
     process_source_file(helper_context, &normalize_path(file_name));
 }
 
 fn get_source_file_from_reference_worker(
     file_name: &str,
-    get_source_file: &mut dyn FnMut(&str) -> Option<Rc<SourceFile>>,
-) -> Option<Rc<SourceFile>> {
+    get_source_file: &mut dyn FnMut(&str) -> Option<Rc<Node>>,
```
helixbass, line 211: Again here maybe this should be generic over the closure type rather than taking a reference to a `dyn` closure type

src/compiler/types.rs
```diff
@@ -41,19 +113,84 @@ impl NodeInterface for Node {
...
+            Node::SourceFile(source_file) => source_file.set_parent(parent),
         }
     }
 }
 
 #[derive(Debug)]
 pub struct BaseNode {
     pub kind: SyntaxKind,
+    pub parent: RwLock<Option<Weak<Node>>>,
+    pub pos: AtomicUsize,
```
helixbass, line 143: I used `AtomicUsize`'s here so that `set_pos()`/`set_end()` could take `&self` rather than `&mut self`, otherwise I would've been in a situation where since the `Node` has already been wrapped in an `Rc` (when we want to set its `pos`/`end`) it doesn't want to give us mutable access to its contents

src/compiler/types.rs
```diff
@@ -41,19 +113,84 @@ impl NodeInterface for Node {
...
+            .upgrade()
+            .unwrap()
+    }
+
+    fn set_parent(&self, parent: Rc<Node>) {
+        *self.parent.write().unwrap() = Some(Rc::downgrade(&parent));
+    }
+}
+
+impl ReadonlyTextRange for BaseNode {
```
helixbass, line 178: This is where it's nice that all of the concrete AST node types eg `PrefixUnaryExpression` are using the same nested `BaseNode` concrete type under the hood for storing the actual `NodeInterface` fields because I only have to implement this atomic stuff once rather than n times

src/compiler/types.rs
```diff
@@ -64,18 +201,43 @@ impl From<BaseNode> for Node {
...
+    _nodes: Vec<Rc<Node>>,
 }
 
 impl NodeArray {
-    pub fn new(nodes: Vec<Node>) -> Self {
+    pub fn new(nodes: Vec<Rc<Node>>) -> Self {
         NodeArray { _nodes: nodes }
     }
+
+    pub fn iter(&self) -> NodeArrayIter {
```
helixbass, line 212: This stuff I got from [this answer](https://stackoverflow.com/a/40668157/732366) for how to implement an iterator on what is basically a `Vec` wrapper

src/compiler/types.rs
```diff
@@ -314,6 +898,255 @@ pub enum ExitStatus {
...
+    _intrinsic_type: BaseIntrinsicType,
+    pub fresh_type: WeakSelf<Type>,
+    pub regular_type: WeakSelf<Type>,
+}
+
+impl FreshableIntrinsicType {
+    pub fn new(intrinsic_type: BaseIntrinsicType) -> Self {
+        Self {
+            _intrinsic_type: intrinsic_type,
+            fresh_type: WeakSelf::new(),
```
helixbass, line 1036: Here's the rest of the `WeakSelf` API that I used, `WeakSelf::new()` initializes it and then below `.get()` returns the expected (already-initialized via `.init()` somewhere) weak reference (or else I guess panics if you haven't actually initialized it yet?)

src/compiler/utilities.rs
```diff
@@ -1,8 +1,58 @@
...
+    DiagnosticRelatedInformationInterface, DiagnosticWithDetachedLocation, DiagnosticWithLocation,
+    Node, NodeInterface, ReadonlyTextRange, SortedArray, SourceFile, SyntaxKind, TextSpan,
+    TypeFlags,
+};
+
+fn get_source_file_of_node<TNode: NodeInterface>(node: &TNode) -> Rc<SourceFile> {
+    if node.kind() == SyntaxKind::SourceFile {
+        unimplemented!()
+    }
+    let mut parent = node.parent();
```
helixbass, line 19: This is where we traverse up the AST (via the `.parent` links initialized by the binder) to get the root `SourceFile` given some AST node

src/compiler/utilities.rs
```diff
@@ -36,39 +86,100 @@ pub fn get_binary_operator_precedence(kind: SyntaxKind) -> OperatorPrecedence {
...
+
+impl DiagnosticCollection {
+    pub fn new() -> Self {
+        DiagnosticCollection {
+            file_diagnostics: HashMap::<String, SortedArray<Rc<Diagnostic>>>::new(),
+        }
+    }
+
+    pub fn add(&mut self, diagnostic: Rc<Diagnostic>) {
+        // let diagnostics = self
```
helixbass, line 101: This is another place where the borrow checker was unhappy with this way of writing the logic

src/rust_helpers/weak_self.rs
```diff
@@ -0,0 +1,79 @@
...
+        WeakSelf {
+            cell: UnsafeCell::new(None),
+        }
+    }
+
+    /// Initialize the WeakSelf<T> with an Rc.
+    ///
+    /// Note: content must point be the only existing Rc, otherwise this method will panic
+    pub fn init(&self, content: &Rc<T>, require_exclusive_access: bool) {
+        if require_exclusive_access {
```
helixbass, line 31: I modified `WeakSelf` to accept a flag for whether to panic if we don't have exclusive access to the `Rc` contents at the time of initialization because for the `true_type`/`regular_true_type` use case we've already initialized one weak reference to each before we can then initialize the other weak reference to them (since they both get referenced by themselves and each other)

src/rust_helpers/weak_self.rs
```diff
@@ -0,0 +1,79 @@
+// copied from https://github.com/eun-ice/weak-self
+
+use std::cell::UnsafeCell;
+use std::fmt;
+use std::rc::{Rc, Weak};
```
helixbass, line 5: I modified `WeakCell` to use `Rc`'s rather than `Arc`'s

src/rust_helpers/weak_self.rs
```diff
@@ -0,0 +1,79 @@
...
+
+pub struct WeakSelf<T: ?Sized> {
+    cell: UnsafeCell<Option<Weak<T>>>,
+}
+
+impl<T: ?Sized> Clone for WeakSelf<T> {
```
helixbass, line 11: And I implemented `Clone` for `WeakSelf`

src/compiler/checker.rs
peterstuart: I like that idea

src/compiler/core.rs
peterstuart: Yep, I've run into the same problem trying to make the inner value of an Option generic. I think I ended up just making a variants of the function rather than using an Option, but that's annoying.

src/compiler/core.rs
peterstuart: This seems like it should be a function on `SortedArray`.

src/compiler/core_public.rs
```diff
@@ -0,0 +1,17 @@
...
+}
+
+impl<TItem> SortedArray<TItem> {
+    pub fn new(vec: Vec<TItem>) -> Self {
+        Self { _vec: vec }
+    }
+
+    pub fn push(&mut self, item: TItem) {
+        self._vec.push(item);
+    }
```
peterstuart, line 12: The functions on `SortedArray` seem like they should enforce that it's sorted. (ie. `new` should sort the incoming `Vec` and `push` should insert the new item at the correct index.)

src/compiler/core_public.rs
```diff
@@ -0,0 +1,17 @@
+pub struct SortedArray<TItem> {
+    pub _vec: Vec<TItem>,
```
peterstuart, line 2: Similar to the comment below, it seems bad to make this public, because anyone could mutate it, breaking the contract that a `SortedArray` is sorted.

src/compiler/parser.rs
peterstuart: Plain `Fn` seems to work rather than `FnMut` for both of these.

src/rust_helpers/weak_self.rs
peterstuart: What does the "?" in front of "Sized" do?

peterstuart (commented):
I'm very lost in the logic, but I left some code-level comments.

src/rust_helpers/weak_self.rs
helixbass: I didn't write this code (copied from existing `weak_self` crate) but I think that by default generics have a `Sized` bound and so `?Sized` is the syntax for "opting out" of that

src/compiler/binder.rs
```diff
@@ -0,0 +1,194 @@
...
+        }
+    }
+
+    fn get_container_flags(&self, node: Rc<Node>) -> ContainerFlags {
+        match node.kind() {
+            SyntaxKind::SourceFile => {
+                return ContainerFlags::IsContainer
+                    | ContainerFlags::IsControlFlowContainer
+                    | ContainerFlags::HasLocals;
+            }
```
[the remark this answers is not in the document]
helixbass: Because the following `ContainerFlags::None` is a following statement in this function body, so if we don't explicitly `return` here then this would just be the value of this `match` expression but that would get silently discarded and then `ContainerFlags::None` would always get returned

src/compiler/core_public.rs
helixbass: Yup they will, so far I haven't implemented any of the actual sorting logic

src/compiler/core_public.rs
helixbass: Good point, updated to make `_vec` private and for the one existing place that was accessing `_vec` (where it's trying to extract (a copy of) the inner `Vec`), I implemented `From<&SortedArray<TItem>> for Vec<TItem>` so that it can do `.into()` instead of `._vec.clone()`

src/compiler/checker.rs
helixbass: Per other comment, there is in fact a `SortedArray` type but for whatever reason the Typescript version of this helper accepts a plain old `Type[]` rather than a `SortedArray` 🤷‍♀️

PR #7, opened 2021-12-19
Multiple statements

helixbass:
In this PR:
- handle newlines/multiple statements

To test:
If you have a file named eg `tmp.tsx` in the repo root directory whose contents are something like (trailing newline is fine):
```
++1; 2;
++true
3
```
and run `cargo run tmp.tsx` you should see debug-logging of the `SourceFile` with four `ExpressionStatement`'s and then a single `DiagnosticWithLocation` whose `start`/`length` should correspond to the position of the `true` literal (in this example, `start: 10, length: 4`)

Based on `plusplus-number`

### DIFF
--- a/src/compiler/binder.rs
+++ b/src/compiler/binder.rs
@@ -48,19 +48,19 @@ impl BinderType {
     }
 
     fn file(&self) -> Rc<Node> {
-        self.file.read().unwrap().as_ref().unwrap().clone()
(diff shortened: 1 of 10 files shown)


helixbass:
> if you add a hash to your println format strings, it'll pretty print them

Nice, I was hoping something like that was "baked in", updated to use `{:#?}`

PR #8, opened 2021-12-21
Parse variable declaration with type annotation

helixbass:
In this PR:
- parse a `const` variable declaration with `: number` type annotation

To test:
If you have a file named eg `tmp.tsx` in the repo root directory whose contents are `const a: number = true` and run `cargo run tmp.tsx` you should see debug-logging of the `SourceFile` with a `VariableStatement` with a `VariableDeclarationList` containing a single `VariableDeclaration` whose `name` is the `Identifier` `"a"`, `initializer` is a `TrueKeyword` and `type_` is a `NumberKeyword` `KeywordTypeNode` and then it should panic during binding


src/compiler/parser.rs
```diff
@@ -674,7 +1039,10 @@ fn Parser() -> ParserType {
 //     ParserMut.lock().unwrap()
 // }
 
-#[derive(Copy, Clone)]
-enum ParsingContext {
-    SourceElements,
+bitflags! {
+    pub struct ParsingContext: u32 {
```
helixbass, line 1043: `ParsingContext` is implemented as a "normal" Typescript enum rather than a "bitflag-ish" enum (where the enum values are specified as `1 << ...`) but it seems to be being used like bitflags (by doing `1 << [enum value]` at the point of usage) so it seems like it makes more sense to just use the bitflags pattern (not sure why they aren't doing that in the Typescript version since they use that pattern for a lot of other enums)

src/compiler/scanner.rs
```diff
@@ -78,21 +81,21 @@ impl Scanner {
...
     }
 
-    fn get_identifier_token(&mut self) -> SyntaxKind {
+    fn get_identifier_token(&self) -> SyntaxKind {
         let len = self.token_value().len();
         if len >= 2 && len <= 12 {
             let ch = self.token_value().chars().nth(0).unwrap();
             if ch >= CharacterCodes::a && ch <= CharacterCodes::z {
-                let keyword = text_to_keyword.get(self.token_value());
+                let keyword = text_to_keyword.get(&self.token_value());
```
helixbass, line 89: This is what caused me to change `text_to_keyword` from `HashMap<&'static str, SyntaxKind>` to `HashMap<String, SyntaxKind>`, for some reason when I changed `self.token_value()` to return a `String` rather than an `&str` (because now that `token_value` is wrapped in an `RwLock` it doesn't want to return an `&str` reference to its contents) it was unhappy with me passing `&self.token_value()` as an argument to `text_to_keyword.get()`

src/compiler/parser.rs
```diff
@@ -91,34 +101,44 @@ impl ReadonlyTextRange for MissingNode {
 
 #[allow(non_snake_case)]
 struct ParserType {
-    scanner: Scanner,
+    scanner: RwLock<Scanner>,
```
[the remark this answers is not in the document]
peterstuart: I've never really thought about using stuff like this in a single-threaded program, but that use makes sense. It seems like there is probably some risk of deadlocks, but that'd be very obvious when you hit it.

src/compiler/parser.rs
```diff
@@ -385,21 +516,70 @@ impl ParserType {
...
+        }
+    }
+
+    fn parse_delimited_list<TItem: Into<Node>>(
+        &mut self,
+        kind: ParsingContext,
+        parse_element: fn(&mut ParserType) -> TItem,
+    ) -> NodeArray {
+        let save_parsing_context = self.parsing_context();
+        self.set_parsing_context(self.parsing_context() | kind);
```
peterstuart, line 553: What does it mean to `|` these things?

src/compiler/parser.rs
```diff
@@ -629,18 +882,130 @@ impl ParserType {
...
+        self.next_token();
+
+        let declarations: NodeArray;
+        if false {
+            unimplemented!()
+        } else {
+            declarations = self.parse_delimited_list(
+                ParsingContext::VariableDeclarations,
+                ParserType::parse_variable_declaration_allow_exclamation,
+            );
```
peterstuart, line 991: I'm not sure I realized you could do this (declare a variable but not set it right away). A maybe slightly more idiomatic way to do that would be to `let declarations = if ...`, but maybe that doesn't work with the future code of that if-statement.

src/compiler/parser.rs
helixbass: The `bitflags` crate automatically implements the bitwise operators for your bitflag types so this is "bitwise-or"'ing these two `ParsingContext` bitflag values (so basically the union of those bitflags)

src/compiler/parser.rs
helixbass: Ya I wasn't sure if Rust supported delayed initialization like this but then ya this mimics the Typescript version pretty closely

PR #9, opened 2021-12-21
Bind variable declaration to symbol

helixbass:
In this PR:
- create symbol table and populate with a symbol corresponding to a variable declaration during binding

To test:
If you have a file named eg `tmp.tsx` in the repo root directory whose contents are `const a: number = true` and run `cargo run tmp.tsx` you should see debug-logging of the (pre-binding) `SourceFile`, then a `post-binding:` debug-logging of the `SourceFile` where the `SourceFile` node's `locals` is a symbol table mapping from the `__String` `"a"` to a `Symbol` with `escaped_name: "a"` and `declarations` a `Vec` containing the `VariableDeclaration` AST node that declared `a`, and the `symbol` on the `BaseNode` inside that `VariableDeclaration` should be a populated weak reference (to the `Symbol`)  and then it should panic during type-checking

Based on `variable-declaration`


src/compiler/binder.rs
```diff
@@ -63,19 +71,113 @@ impl BinderType {
...
+        self.set_container(None);
+    }
+
+    fn create_symbol(&self, flags: SymbolFlags, name: __String) -> Symbol {
+        self.Symbol()(flags, name)
+    }
+
+    fn add_declaration_to_symbol(&self, symbol: Rc<Symbol>, node: Rc<Node /*Declaration*/>) {
+        node.set_symbol(symbol.clone());
+        let declarations = {
```
helixbass, line 120: The block scoping is to ensure that the `RwLockReadGuard` returned by `symbol.maybe_declarations()` is dropped before we call `.set_declarations()` (which will want write access to the `RwLock`)

This whole pattern is a little wonky from a Rust perspective - `appendIfUnique()` mutates the passed array if it exists or else returns a new array, but since I can't give it ownership of the underlying existing `symbol.declarations` `Vec` it'd have to take an `&mut` to the `Vec` in order to avoid creating a new `Vec`, but that's not compatible with the passed-`None`-and-creates-new-`Vec` use case (since it'd have to return a full-fledged new `Vec`) (is this somewhere I could use `Cow`?)?

src/compiler/binder.rs
```diff
@@ -63,19 +71,113 @@ impl BinderType {
...
         self.set_parent(None);
+        self.set_container(None);
+    }
+
+    fn create_symbol(&self, flags: SymbolFlags, name: __String) -> Symbol {
+        self.Symbol()(flags, name)
+    }
+
+    fn add_declaration_to_symbol(&self, symbol: Rc<Symbol>, node: Rc<Node /*Declaration*/>) {
+        node.set_symbol(symbol.clone());
```
helixbass, line 119: `Symbol`'s get stored in a symbol table but also as a field on `Node`'s, so I'm wrapping them in `Rc`'s and storing a weak reference in `BaseNode`'s `symbol` field

src/compiler/binder.rs
```diff
@@ -63,19 +71,113 @@ impl BinderType {
...
+    ) -> Rc<Symbol> {
+        let name = self.get_declaration_name(node.clone());
+
+        let mut symbol = None;
+        match name {
+            None => unimplemented!(),
+            Some(name) => {
+                if true {
+                    symbol = Some(Rc::new(self.create_symbol(SymbolFlags::None, name.clone())));
+                    symbol_table.insert(name, symbol.as_ref().unwrap().clone());
```
helixbass, line 154: This is the first time I'd used type aliasing in Rust I think, pretty slick how I can just call methods like `.insert()` (`SymbolTable` is aliased to a certain type of `HashMap`)

src/compiler/parser.rs
```diff
@@ -22,25 +25,58 @@ enum SpeculationKind {
...
-    cb_node: TNodeCallback,
+    mut cb_node: TNodeCallback,
     cb_nodes: TNodesCallback,
 ) {
     if node.kind() <= SyntaxKind::LastToken {
         return;
     }
     match &*node {
+        Node::VariableDeclaration(variable_declaration) => {
+            visit_node(&mut cb_node, Some(variable_declaration.name()));
```
helixbass, line 58: The compiler was complaining that this call of `visit_node()` takes ownership of the `cb_node` closure argument (and then I try and call `visit_node()` with it again right after this), which I guess makes sense, so I modified `visit_node()` to take a reference to a closure instead of a closure

src/compiler/utilities.rs
```diff
@@ -54,6 +60,24 @@ fn get_error_span_for_node<TNode: NodeInterface>(
...
+        | SyntaxKind::StringLiteral
+        | SyntaxKind::NoSubstitutionTemplateLiteral
+        | SyntaxKind::NumericLiteral => true,
+        _ => false,
+    }
+}
+
+pub fn get_escaped_text_of_identifier_or_literal(node: Rc<Node>) -> __String {
+    if is_member_name(&*node) {
+        node.as_member_name().escaped_text()
```
helixbass, line 75: I like this `node.as_*()` pattern, eg I exposed a `.as_member_name()` method on `Node` that panics unless it can unwrap the `Node` into one of the AST node sub-variants that implements `MemberNameInterface` (and it returns a `&dyn MemberNameInterface`). It's a little bit like what I was thinking of doing via a macro (eg `unwrap_ast_node!(node, BinaryExpression)`) for unwrapping to concrete AST node types but for unwrapping to a certain AST node interface instead

src/compiler/types.rs
```diff
@@ -203,6 +302,8 @@ pub struct BaseNode {
     pub parent: RwLock<Option<Weak<Node>>>,
     pub pos: AtomicUsize,
     pub end: AtomicUsize,
+    pub symbol: RefCell<Option<Weak<Symbol>>>,
```
helixbass, line 305: I started wondering why I was using `RwLock` instead of `RefCell` (since it seems like you see `Rc<RefCell<...>>`'s more than `Rc<RwLock<...>>`'s). I'm not totally sure but I think I can replace a lot of the `RwLock`'s with `RefCell`'s and that per eg [this article](https://manishearth.github.io/blog/2015/05/27/wrapper-types-in-rust-choosing-your-guarantees/) `RefCell`'s sound cheaper/faster

src/compiler/types.rs
```diff
@@ -234,6 +337,24 @@ impl NodeInterface for BaseNode {
...
+    fn symbol(&self) -> Rc<Symbol> {
+        self.symbol.borrow().as_ref().unwrap().upgrade().unwrap()
+    }
+
+    fn set_symbol(&self, symbol: Rc<Symbol>) {
+        *self.symbol.borrow_mut() = Some(Rc::downgrade(&symbol));
+    }
+
+    fn locals(&self) -> MappedRwLockWriteGuard<SymbolTable> {
+        RwLockWriteGuard::map(self.locals.try_write().unwrap(), |option| {
```
helixbass, line 350: Here's a usage of the `RwLockWriteGuard::map()` that `parking_lot` provides

src/compiler/binder.rs
peterstuart: Hmm, I don't really have any better ideas than what you've done.

src/compiler/binder.rs
```diff
@@ -63,19 +71,113 @@ impl BinderType {
...
+
+    fn get_declaration_name(&self, node: Rc<Node>) -> Option<__String> {
+        let name = get_name_of_declaration(node);
+        if let Some(name) = name {
+            return if is_property_name_literal(&*name) {
+                Some(get_escaped_text_of_identifier_or_literal(name.clone()))
+            } else {
+                None
+            };
+        }
```
peterstuart, line 137: This looks like it could be expressed with `and_then` (flat map) instead of an if-statement, but maybe isn't compatible with whatever will replace `unimplemented!`.

src/compiler/core.rs
```diff
@@ -37,6 +37,24 @@ pub fn insert_sorted<TItem>(array: &mut SortedArray<TItem>, insert: TItem) {
...
+    }
+}
+
+pub fn append_if_unique<TItem>(array: Option<Vec<TItem>>, to_add: TItem) -> Vec<TItem> {
+    if let Some(mut array) = array {
+        push_if_unique(&mut array, to_add);
+        array
+    } else {
+        vec![to_add]
+    }
```
peterstuart, line 55: I think this could be written using one of the `map_or*` functions.

src/compiler/types.rs
peterstuart: Ah ok, yeah I had no idea that `RefCell` was just the single-threaded version of `RWLock`. (The naming doesn't really help there.)

src/compiler/binder.rs
helixbass: Does a `return` inside a closure body return from the enclosing function (like a Ruby block) or just from the closure (relevant here because this block that would go inside `.and_then()` needs to return from the function)?

peterstuart: Just from the closure

PR #10, opened 2021-12-23
ast_type macro: generate basic node interface impl's

helixbass:
In this PR:
- create `local-macros` sub-crate for defining macros to be consumed by the main codebase
- expose `#[ast_type]` macro which when applied to a "leaf" AST node type generates `impl NodeInterface` and `impl ReadonlyTextRange` for that node type

To test:
Things should still behave the same as of #9


local-macros/src/lib.rs
```diff
@@ -0,0 +1,88 @@
+use proc_macro::TokenStream;
+use quote::quote;
+use syn::Data::Struct;
+use syn::{parse_macro_input, DeriveInput, Fields, FieldsNamed};
+
+#[proc_macro_attribute]
+pub fn ast_type(_attr: TokenStream, item: TokenStream) -> TokenStream {
+    let item_for_parsing = item.clone();
+    let DeriveInput {
```
helixbass, line 9: Technically I think `DeriveInput` is supposed to be for when you're implementing a `derive` macro but it seems to work here too (and I didn't see another way mentioned for parsing the macro input (with `syn`) in a way that the attribute could be being applied to either a struct or an enum - for now it's just structs but the non-leaf AST node types (eg `Expression`) are enums and I'd like to be able to apply the same macro to those too)

local-macros/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+                    .expect("Expected at least one struct field")
+                    .ident
+                    .clone(),
+                _ => panic!("Expected named fields"),
+            };
+
+            quote! {
+                impl crate::NodeInterface for #ast_type_name {
+                    fn kind(&self) -> crate::SyntaxKind {
+                        self.#first_field_name.kind()
```
helixbass, line 30: This is assuming the convention where the first struct field is where the nested `BaseNode` lives so we're just forwarding these `NodeInterface` methods on to that field

local-macros/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+                    fn set_end(&self, end: usize) {
+                        self.#first_field_name.set_end(end);
+                    }
+                }
+            }
+        }
+        _ => panic!("Expected struct"),
+    };
+
+    let item_as_proc_macro2_token_stream = proc_macro2::TokenStream::from(item);
```
helixbass, line 80: I guess `quote` expects `proc_macro2` `TokenStream`'s (rather than the built-in `proc_macro` `TokenStream`'s) so I had to convert it in order for this to work (not sure what the typical pattern looks like for emitting the original item input unchanged but also emitting new stuff based on having parsed it?)

peterstuart (approved):
Nice, this is way less complicated than I was expecting.

local-macros/src/lib.rs
peterstuart: Is there a reason this isn't a `derive` macro?

helixbass: Ya I'm probably going to add additional "parameters" to it eg `#[ast_node(interfaces = [MemberNameInterface])`

PR #11, opened 2021-12-23
ast_type macro: generate basic node interface impl's for enum types

helixbass:
In this PR:
- generate `impl NodeInterface` and `impl ReadonlyTextRange` when `#[ast_type]` is applied to a non-leaf (enum) AST node type (eg `Expression`)

To test:
Things should still behave the same as of #9 
If you're curious to see the output of the macros you can `cargo install cargo-expand` and run eg `cargo expand --lib compiler::types > expanded` (from the repo root) and inspect the contents of the `expanded` file

Based on `ast-type-macro-node-interface`


local-macros/src/lib.rs
```diff
@@ -74,15 +74,100 @@ pub fn ast_type(_attr: TokenStream, item: TokenStream) -> TokenStream {
                 }
             }
         }
-        _ => panic!("Expected struct"),
+        Enum(DataEnum { variants, .. }) => {
+            let variant_names = variants.iter().map(|variant| &variant.ident);
+            let variant_names_2 = variant_names.clone();
```
helixbass, line 79: Apparently `quote!` is taking ownership of `variant_names` when I use it in a `#(...#variant_names...),*` so I made n copies of it

local-macros/src/lib.rs
```diff
@@ -74,15 +74,100 @@ pub fn ast_type(_attr: TokenStream, item: TokenStream) -> TokenStream {
...
+            let variant_names_8 = variant_names.clone();
+            let variant_names_9 = variant_names.clone();
+            let variant_names_10 = variant_names.clone();
+            let variant_names_11 = variant_names.clone();
+
+            quote! {
+                impl crate::NodeInterface for #ast_type_name {
+                    fn kind(&self) -> crate::SyntaxKind {
+                        match self {
+                            #(#ast_type_name::#variant_names(nested) => nested.kind()),*
```
helixbass, line 94: This is the `quote` syntax for looping through an iterable

peterstuart: Is there any way to iterate over references (to avoid all the copies of `variant_names`?).

peterstuart (approved):

local-macros/src/lib.rs
helixbass: Ya I'm not sure, didn't see anything to that effect in glancing at the `quote` docs

PR #12, opened 2021-12-23
Use RefCell/Cell's instead of RwLock/atomic's

helixbass:
In this PR:
- instead of using `RwLock`/`Atomic*`, use `RefCell`/`Cell` since it's apparently cheaper/faster

To test:
Should still behave the same as of #9 

Based on `ast-type-macro-node-enum-types`

### DIFF
--- a/src/compiler/parser.rs
+++ b/src/compiler/parser.rs
@@ -1,10 +1,8 @@
 #![allow(non_upper_case_globals)]
 
 use bitflags::bitflags;
-use std::cell::RefMut;
+use std::cell::{Cell, Ref, RefCell, RefMut};
 use std::rc::Rc;
-use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::{RwLock, RwLockReadGuard, RwLockWriteGuard};
 
 use crate::{
     create_detached_diagnostic, create_node_factory, create_scanner,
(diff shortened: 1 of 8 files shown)


src/compiler/parser.rs
```diff
@@ -161,42 +159,42 @@ impl ReadonlyTextRange for MissingNode {
...
     SourceFileConstructor: Option<fn(SyntaxKind, usize, usize) -> BaseNode>,
     factory: NodeFactory,
     file_name: Option<String>,
-    parse_diagnostics: Option<RwLock<Vec<DiagnosticWithDetachedLocation>>>,
-    current_token: RwLock<Option<SyntaxKind>>,
+    parse_diagnostics: Option<RefCell<Vec<DiagnosticWithDetachedLocation>>>,
+    current_token: RefCell<Option<SyntaxKind>>,
     parsing_context: Option<ParsingContext>,
-    parse_error_before_next_finished_node: AtomicBool,
+    parse_error_before_next_finished_node: Cell<bool>,
```
helixbass, line 172: I guess you can use `Cell`'s instead of `RefCell`'s when the wrapped data is `Copy`

src/compiler/types.rs
```diff
@@ -184,32 +176,30 @@ impl NodeInterface for BaseNode {
         *self.symbol.borrow_mut() = Some(Rc::downgrade(&symbol));
     }
 
-    fn locals(&self) -> MappedRwLockWriteGuard<SymbolTable> {
-        RwLockWriteGuard::map(self.locals.try_write().unwrap(), |option| {
-            option.as_mut().unwrap()
-        })
+    fn locals(&self) -> RefMut<SymbolTable> {
+        RefMut::map(self.locals.borrow_mut(), |option| option.as_mut().unwrap())
```
helixbass, line 180: `RefCell`'s support `map()`'ing the borrowed contents, `Ref::map()`/`RefMut::map()` just return the same type as `.borrow()`/`.borrow_mut()` (`Ref`/`RefMut`)

src/compiler/types.rs
```diff
@@ -184,32 +176,30 @@ impl NodeInterface for BaseNode {
...
     fn set_locals(&self, locals: SymbolTable) {
-        *self.locals.try_write().unwrap() = Some(locals);
+        *self.locals.borrow_mut() = Some(locals);
     }
 }
 
 impl ReadonlyTextRange for BaseNode {
     fn pos(&self) -> usize {
-        self.pos.load(Ordering::Relaxed)
+        self.pos.get()
```
helixbass, line 190: `.get()`/`.set()` is the API for reading/writing `Cell`'s

peterstuart (commented):
Faster and also more concise. Can't beat that 🥳

PR #13, opened 2021-12-24
Store a reference to Rc<Node> wrapper in AST nodes

helixbass:
The starting point here is basically option (c) from [this comment](https://github.com/helixbass/tsc-rust/pull/5#discussion_r771830998)

In this PR:
- add a `_node_wrapper: RefCell<Option<Weak<Node>>>` field to the shared `BaseNode` struct for holding a weak reference to the "wrapper" `Rc<Node>` to which it belongs, and add `node_wrapper()`/`set_node_wrapper()` to `NodeInterface` (so that effectively you can get a reference to your wrapping `Rc<Node>` from any AST node type/sub-variant)
- to wire up setting that node wrapper reference upon node creation, make the `#[ast_node]` macro generate an `impl From<[node type]> for Rc<Node>` (along with other existing boilerplate "wrapping" `impl From`'s) that encapsulates setting that node wrapper right after it's created (and then always use `.into()` for creating `Rc<Node>`'s during parsing)
- with that in place, stop passing around an additional `wrapper` argument when access to the `Rc<Node>` wrapper is needed, and make a lot of functions not take `Rc<Node>`'s that don't really care about its "`Rc`-ness" (since we can now obtain an `Rc<Node>` wrapper reference from anything implementing `NodeInterface`)

To test:
Things should still behave the same as of #7 (if you do the testing from #9 it'd basically do the same thing as described there but would panic in a slightly different way because of the in-progress work from `variable-declaration-type-checking` (see below))

Based on `variable-declaration-type-checking` (a branch I've been working on but haven't gotten up for review yet)

### DIFF
--- a/local-macros/src/lib.rs
+++ b/local-macros/src/lib.rs
@@ -1,10 +1,41 @@
+use darling::FromMeta;
 use proc_macro::TokenStream;
+use proc_macro2::Ident;
 use quote::quote;
 use syn::Data::{Enum, Struct};
-use syn::{parse_macro_input, DataEnum, DeriveInput, Fields, FieldsNamed};
+use syn::{parse_macro_input, AttributeArgs, DataEnum, DeriveInput, Fields, FieldsNamed};
+
+#[derive(Debug, FromMeta)]
+struct AstTypeArgs {
+    #[darling(default)]
+    ancestors: Option<String>,
+    #[darling(default)]
+    impl_from: Option<bool>,
+}
+
+impl AstTypeArgs {
+    fn ancestors_vec(&self) -> Vec<String> {
+        let mut vec = self.ancestors.as_ref().map_or_else(
+            || vec![],
+            |ancestors_str| {
+                ancestors_str
+                    .split(",")
+                    .into_iter()
+                    .map(|chunk| chunk.trim().to_string())
+                    .collect()
+            },
+        );
+        vec.push("Node".to_string());
+        vec
+    }
+
+    fn should_impl_from(&self) -> bool {
+        self.impl_from.unwrap_or(true)
+    }
+}
 
 #[proc_macro_attribute]
-pub fn ast_type(_attr: TokenStream, item: TokenStream) -> TokenStream {
+pub fn ast_type(attr: TokenStream, item: TokenStream) -> TokenStream {
     let item_for_parsing = item.clone();
     let DeriveInput {
         ident: ast_type_name,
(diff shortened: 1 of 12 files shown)


src/compiler/parser.rs
```diff
@@ -512,8 +525,12 @@ impl ParserType {
     }
 
     fn create_node_array(&self, elements: Vec<Node>) -> NodeArray {
-        self.factory
-            .create_node_array(elements.into_iter().map(Rc::new).collect::<Vec<Rc<Node>>>())
+        self.factory.create_node_array(
+            elements
+                .into_iter()
+                .map(Node::wrap)
```
helixbass, line 531: Here's somewhere where I needed to convert `Node` -> `Rc<Node>` - I couldn't generate an `impl From<Node> for Rc<Node>` because there's already a (conflicting) blanket implementation for that, so I hand-rolled a `Node::wrap()` method to do the equivalent thing (ie encapsulate the setting of the weak reference to the node wrapper at the time of its creation)

src/compiler/program.rs
```diff
@@ -235,7 +235,7 @@ fn find_source_file_worker(
     let file = helper_context.host.get_source_file(file_name);
 
     file.map(|file| {
-        let file: Rc<Node> = Rc::new(Rc::new(file).into());
+        let file: Rc<Node> = file.into();
```
helixbass, line 238: I couldn't implement `impl From<Rc<SourceFile>> for Rc<Node>` because I don't own the `Rc` type (so I was violating the "orphan rule"), so I adjusted this to be a `From<SourceFile> for Rc<Node>` instead (which knows that `SourceFile` is a special case where it gets its own direct `Rc<SourceFile>` wrapper before being wrapped in an `Rc<Node>`)

local-macros/src/lib.rs
```diff
@@ -1,10 +1,41 @@
+use darling::FromMeta;
 use proc_macro::TokenStream;
+use proc_macro2::Ident;
 use quote::quote;
 use syn::Data::{Enum, Struct};
-use syn::{parse_macro_input, DataEnum, DeriveInput, Fields, FieldsNamed};
+use syn::{parse_macro_input, AttributeArgs, DataEnum, DeriveInput, Fields, FieldsNamed};
+
+#[derive(Debug, FromMeta)]
```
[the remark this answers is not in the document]
peterstuart: > #[ast_node(ancestors = "LiteralLikeExpression, Expression")]

I'd have expected this to use an array:

```
#[ast_node(ancestors = ["LiteralLikeExpression", "Expression"])]
```

(I don't know if that complicated things?)

local-macros/src/lib.rs
```diff
@@ -91,9 +130,23 @@ pub fn ast_type(_attr: TokenStream, item: TokenStream) -> TokenStream {
             let variant_names_10 = variant_names.clone();
             let variant_names_11 = variant_names.clone();
             let variant_names_12 = variant_names.clone();
+            let variant_names_13 = variant_names.clone();
+            let variant_names_14 = variant_names.clone();
```
peterstuart, line 134: If you `collect` `variant_names` into a `Vec`, you can reuse it and the clones aren't necessary.

src/compiler/binder.rs
```diff
@@ -310,21 +309,20 @@ impl BinderType {
...
 
-    fn bind(&self, node: Option<Rc<Node>>) {
-        let node = match node.as_ref() {
-            None => {
-                return;
-            }
-            Some(node) => node.clone(),
-        };
-        set_parent(&*node, self.maybe_parent());
+    fn bind<TNodeRef: Borrow<Node>>(&self, node: Option<TNodeRef>) {
```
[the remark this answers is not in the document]
peterstuart: Ah cool, I hadn't' seen `Borrow` before.

peterstuart (commented):
👍🏻 Looks like a nice cleanup.

local-macros/src/lib.rs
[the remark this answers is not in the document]
helixbass: Per [this comment](https://www.reddit.com/r/rust/comments/j6afjj/comment/g7zhdmf/?utm_source=share&utm_medium=web2x&context=3) it sounds like I wouldn't be able to leverage `syn` if I wanted to use the more intuitive "array" syntax like you're describing

local-macros/src/lib.rs
helixbass: Per a later PR I ended up getting rid of these by using a reference per an earlier comment of yours

Ah actually I think you're pointing at the real reason it started working (because I believe I updated it to be a reference to a `Vec`), if it was an iterator before then it would've gotten consumed by a single usage but a `Vec` can be reused?

peterstuart: Yeah, I think that's it.

PR #14, opened 2021-12-24
Generate other node interfaces

helixbass:
In this PR:
- generate other node interfaces via `#[ast_node(interfaces = "...")]`

To test:
Behavior should still be the same as of previous PR

Based on `get-node-wrapper`

### DIFF
--- a/local-macros/src/lib.rs
+++ b/local-macros/src/lib.rs
@@ -37,16 +37,15 @@ impl AstTypeArgs {
 
     fn interfaces_vec(&self) -> Vec<String> {
         let mut vec = vec!["ReadonlyTextRange".to_string(), "NodeInterface".to_string()];
-        vec.append(&mut self.interfaces.as_ref().map_or_else(
-            || vec![],
-            |interfaces_str| {
-                interfaces_str
+        if let Some(interfaces_str) = self.interfaces.as_ref() {
+            vec.append(
+                &mut interfaces_str
                     .split(",")
                     .into_iter()
                     .map(|chunk| chunk.trim().to_string())
-                    .collect()
-            },
-        ));
+                    .collect(),
+            );
+        }
         vec
     }
 }
--- a/local-macros/src/lib.rs
+++ b/local-macros/src/lib.rs
@@ -1,6 +1,6 @@
 use darling::FromMeta;
 use proc_macro::TokenStream;
-use proc_macro2::Ident;
+use proc_macro2::{Ident, TokenStream as TokenStream2};
 use quote::quote;
 use syn::Data::{Enum, Struct};
 use syn::{parse_macro_input, AttributeArgs, DataEnum, DeriveInput, Fields, FieldsNamed};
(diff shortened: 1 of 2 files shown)


local-macros/src/lib.rs
```diff
@@ -117,33 +123,86 @@ pub fn ast_type(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+    ast_type_name: &Ident,
+) -> TokenStream2 {
+    match interface_name {
+        "NodeInterface" => {
             quote! {
                 impl crate::NodeInterface for #ast_type_name {
                     fn node_wrapper(&self) -> ::std::rc::Rc<crate::Node> {
                         match self {
-                            #(#ast_type_name::#variant_names_13(nested) => nested.node_wrapper()),*
+                            #(#ast_type_name::#variant_names(nested) => nested.node_wrapper()),*
```
helixbass, line 199: I see now what you meant [here](https://github.com/helixbass/tsc-rust/pull/11#discussion_r774780142), I didn't run into the `#variant_names_n` issue when the type of `variant_names` was a reference

local-macros/src/lib.rs
```diff
@@ -32,29 +34,30 @@ impl AstTypeArgs {
...
-                    .nth(0)
-                    .expect("Expected at least one struct field")
-                    .ident
-                    .clone(),
-                _ => panic!("Expected named fields"),
-            };
+    fn interfaces_vec(&self) -> Vec<String> {
+        let mut vec = vec!["ReadonlyTextRange".to_string(), "NodeInterface".to_string()];
+        vec.append(&mut self.interfaces.as_ref().map_or_else(
+            || vec![],
```
peterstuart, line 41: Since appending an empty vec is a no-op, I'd find this a bit more clear if it were written as an `if let Some(...)` rather than `map_or_else`.

local-macros/src/lib.rs
```diff
@@ -117,33 +123,86 @@ pub fn ast_type(attr: TokenStream, item: TokenStream) -> TokenStream {
...
-            let variant_names_9 = variant_names.clone();
-            let variant_names_10 = variant_names.clone();
-            let variant_names_11 = variant_names.clone();
-            let variant_names_12 = variant_names.clone();
-            let variant_names_13 = variant_names.clone();
-            let variant_names_14 = variant_names.clone();
+        "NamedDeclarationInterface" => {
+            quote! {
+                impl crate::NamedDeclarationInterface for #ast_type_name {
+                    fn name(&self) -> ::std::rc::Rc<crate::Node> {
```
peterstuart, line 129: What does the `::` in front of `std` mean?

local-macros/src/lib.rs
helixbass: Sure, updated to use `if let Some(...)` instead

local-macros/src/lib.rs
helixbass: I think it's sort of like a leading `/` in an absolute path, it's saying "this is a root name, not a relative/local name" (or whatever the right terminology is)

Like in Rails if you were to say `::ModelName` to disambiguate the model from a module-local class named `Module::ModelName`

PR #15, opened 2021-12-25
Variable declaration type checking (part 1)

helixbass:
In this PR:
- part of type-checking for a variable declaration like `const a: number = true`

I'll add testing instructions to part 2 of this PR which is where I "landed" this branch (it's split up into two PR's because I based some of the `ast_type` macro branches on this branch and then based part 2 on top of those)


src/compiler/binder.rs
```diff
@@ -115,15 +115,26 @@ impl BinderType {
...
             append_if_unique(
                 symbol.maybe_declarations().as_ref().map(|vec| vec.clone()),
-                node,
+                node.clone(),
             )
         };
         symbol.set_declarations(declarations);
+
+        if symbol_flags.intersects(SymbolFlags::Value) {
+            set_value_declaration(&*symbol, node);
```
helixbass, line 136: `symbol.value_declaration` is going to be used by the type-checking so wiring that up here in the binder

src/compiler/checker.rs
```diff
@@ -468,6 +733,66 @@ impl TypeChecker {
...
+            )
+        {
+            let initializer = get_effective_initializer(&*wrapper);
+            if let Some(initializer) = initializer {
+                if true {
+                    let initializer_type = self.check_expression_cached(match &*initializer {
+                        Node::Expression(expression) => expression,
+                        _ => panic!("Expected Expression"),
+                    });
+                    self.check_type_assignable_to_and_optionally_elaborate(
```
helixbass, line 762: This is where it's actually checking whether the declared type of the variable is compatible with the type of its initializer

src/compiler/checker.rs
```diff
@@ -170,6 +182,109 @@ impl TypeChecker {
...
+        unimplemented!()
+    }
+
+    fn try_get_type_from_effective_type_node(
+        &self,
+        declaration: &Node, /*Declaration*/
+    ) -> Option<Rc<Type>> {
+        let type_node = get_effective_type_annotation_node(declaration);
+        type_node.map(|type_node| self.get_type_from_type_node(&*type_node))
+    }
```
peterstuart, line 243: I don't know if it's more clear, but you could write this function using the `?` option syntax:

```rs
fn try_get_type_from_effective_type_node(
    &self,
    declaration: &Node, /*Declaration*/
) -> Option<Rc<Type>> {
    let type_node = get_effective_type_annotation_node(declaration)?;
    Some(self.get_type_from_type_node(&*type_node))
}
```

src/compiler/types.rs
```diff
@@ -1389,6 +1495,12 @@ pub struct DiagnosticMessage {
     pub message: &'static str,
 }
 
+#[derive(Clone, Debug)]
+pub struct DiagnosticMessageChain {
+    pub message_text: String,
+    pub next: Option<Vec<DiagnosticMessageChain>>,
```
peterstuart, line 1501: This is probably mimicking the TS code, but seems weird to have an option of a vec (which can be empty), unless empty vec and `None` mean different things.

src/compiler/utilities.rs
```diff
@@ -162,6 +200,13 @@ impl DiagnosticCollection {
     }
 }
 
+pub fn get_effective_type_annotation_node(node: &Node) -> Option<Rc<Node /*TypeNode*/>> {
+    let type_ = node
+        .maybe_as_has_type()
+        .and_then(|has_type| has_type.type_());
+    type_
```
peterstuart, line 207: This is another place where you could use `?` and maybe make this a bit more concise.

src/compiler/checker.rs
helixbass: Ah interesting, I forgot that you can use `?` for `Option`'s in addition to `Result`'s

PR #16, opened 2021-12-25
Variable declaration type checking (part 2)

helixbass:
In this PR:
- Part 2 of type-checking for a variable declaration like `const a: number = true`

To test:
If you run `cargo run tmp.tsx` against a file whose contents are `const a: number = true` you should see a diagnostic debug-logged with `message_text: "Type 'true' is not assignable to type 'number'."`
If you run `cargo run tmp.tsx` against a file whose contents are eg `const a: number = 1` you should see no diagnostics debug-logged
If you follow the testing instructions from #7 you should now see the `message_text: "An arithmetic operand must be of type 'any', 'number', 'bigint' or an enum type."` as part of the debug-logged diagnostic

Based on `generate-other-node-interfaces`


local-macros/src/lib.rs
```diff
@@ -105,19 +105,19 @@ fn get_struct_interface_impl(
         "ReadonlyTextRange" => {
             quote! {
                 impl crate::ReadonlyTextRange for #ast_type_name {
-                    fn pos(&self) -> usize {
+                    fn pos(&self) -> isize {
```
helixbass, line 108: One of the factories was initializing nodes with "dummy" locations of `-1` so converted all of these from `usize` -> `isize`

src/compiler/checker.rs
```diff
@@ -100,13 +106,52 @@ pub fn create_type_checker<TTypeCheckerHost: TypeCheckerHost>(
         _ => panic!("Expected IntrinsicType"),
     }
     type_checker.regular_true_type = Some(regular_true_type);
+    let false_type: Rc<Type> = Rc::new(
```
helixbass, line 109: This is mimicking the pattern with `true_type`/`regular_true_type` above

src/compiler/checker.rs
```diff
@@ -179,12 +240,31 @@ impl TypeChecker {
...
     fn create_intrinsic_type(&self, kind: TypeFlags, intrinsic_name: &str) -> BaseIntrinsicType {
         let type_ = self.create_type(kind);
-        let type_ = BaseIntrinsicType::new(type_);
+        let type_ = BaseIntrinsicType::new(type_, intrinsic_name.to_string());
         type_
     }
 
     fn type_to_string(&self, type_: &Type) -> String {
-        unimplemented!()
+        let writer = Rc::new(RefCell::new(create_text_writer("")));
```
helixbass, line 248: This was sort of messy/confusing, I need to pass this writer to `printer.write_node()` below "by reference" and then read its value (`.get_text()`) after that. So the printer can't take ownership of the writer. I tried passing a reference to the writer to the printer but wasn't getting that to work partly because the writer argument to `.write_node()` (which gets stored in the "state" of the printer) is of a trait type (`EmitTextWriter`?) so seemingly has to get passed/stored as a `dyn` trait object (I forget exactly what I ran into that wasn't working)

So ended up making the type it's stored as on the printer an `Rc<RefCell<...>>` but then here the caller has to be aware of that since the `Rc`-ness is the way that it's able to pass it "by reference" and still use it after

src/compiler/checker.rs
```diff
@@ -816,7 +896,74 @@ impl TypeChecker {
...
+            return Into::<KeywordTypeNode>::into(
+                factory.create_keyword_type_node(&synthetic_factory, SyntaxKind::NumberKeyword),
+            )
+            .into();
+        }
+        if type_.flags().intersects(TypeFlags::BooleanLiteral) {
+            return factory
+                .create_literal_type_node(
+                    &synthetic_factory,
+                    &*Into::<Rc<Node>>::into(
```
helixbass, line 942: Here I ran into it later trying to unwrap these nodes to their `Rc<Node>` wrapper where that wrapper didn't exist since these are "artificially constructed" AST nodes (for error display of eg `'number'`) rather than parsed AST nodes. So had to make sure to wrap them before passing them along

src/compiler/checker.rs
```diff
@@ -880,12 +1027,10 @@ impl<'type_checker> CheckTypeRelatedTo<'type_checker> {
         result != Ternary::False
     }
 
-    fn report_error(&self, message: &DiagnosticMessage) {
+    fn report_error(&self, message: &DiagnosticMessage, args: Option<Vec<String>>) {
```
helixbass, line 1030: I represented these "error format-string arguments" (so eg the strings `true` and `number` that will be plugged into the diagnostic message `"Type '{0}' is not assignable to type '{1}'."`) (that are represented in the Typescript version as either `arg0, arg1, [...]` or variadic `...args`) as an `Option<Vec<String>>` (should probably be `Option<Vec<&str>>`)

src/compiler/checker.rs
```diff
@@ -880,12 +1027,10 @@ impl<'type_checker> CheckTypeRelatedTo<'type_checker> {
...
         Debug_.assert(self.error_node.is_some(), None);
-        self.set_error_info(chain_diagnostic_messages(
-            self.error_info().clone(),
-            message,
-        ));
+        let error_info = { chain_diagnostic_messages(self.error_info().clone(), message, args) };
```
helixbass, line 1032: Added an artificial block scope here to avoid conflicting borrows of `self.error_info`

src/compiler/checker.rs
```diff
@@ -1004,23 +1151,72 @@ impl<'type_checker> CheckTypeRelatedTo<'type_checker> {
...
+                source_type.clone(),
+                target.clone(),
+                report_errors,
+                None,
+                Some(intersection_state),
+            );
+            if related == Ternary::False {
+                return Ternary::False;
+            }
+            result &= related;
```
helixbass, line 1174: The Typescript definition of the `Ternary` enum (where the enum variant values are `0`, `1`, `3` and `-1`) explains how they have those values so that bitwise-`&`'ing/`|`'ing them together gives you the greater or lesser of the two (I don't understand how `-1` works in terms of bitwise operations). I did a little sanity-checking that the same seems to hold true with Rust `i32`'s and then implemented `BitAndAssign` for `Ternary`

src/compiler/emitter.rs
```diff
@@ -0,0 +1,135 @@
...
+    NodeInterface, Printer, PrinterOptions, SourceFile, TypeNode,
+};
+
+#[derive(PartialEq, Eq)]
+enum PipelinePhase {
+    Notification,
+    Emit,
+}
+
+pub fn create_printer(printer_options: PrinterOptions) -> Printer {
```
helixbass, line 15: I guess the `typeToString()` function calls into the emitting code to "print" a type as a string so implemented a couple code paths for emitting

src/compiler/factory/node_factory.rs
```diff
@@ -240,3 +273,57 @@ impl NodeFactory {
 pub fn create_node_factory() -> NodeFactory {
     NodeFactory {}
 }
+
+// lazy_static! {
+//     static ref base_factory_static: BaseNodeFactoryConcrete = create_base_node_factory();
```
helixbass, line 278: Because I need the interface to `BaseNodeFactory` to take immutable `&self` references (because `ParserType` implements `BaseNodeFactory` and needed to pass immutable rather than mutable references to itself into the node factory methods), `BaseNodeFactoryConcrete` which lazily initializes its "node constructor" state has to use `RefCell`'s under the hood which don't meet the `Sync` or whatever requirements for `lazy_static!`

So similar to what I've done in a couple other places, rather than exposing a static singleton I expose a "creator helper" (in this case `create_base_node_factory()`)

The same thing "infected" `BaseNodeFactorySynthetic` below I believe so similarly exposed a `get_synthetic_factory()` "creator helper" rather than a `synthetic_factory` static singleton

src/compiler/parser.rs
```diff
@@ -355,13 +356,13 @@ impl ParserType {
 
     fn parse_error_at_current_token(&self, message: &DiagnosticMessage) {
         self.parse_error_at(
-            self.scanner().get_token_pos(),
-            self.scanner().get_text_pos(),
+            self.scanner().get_token_pos().try_into().unwrap(),
```
helixbass, line 359: The scanner state stayed as `usize`'s (rather than also converting to `isize`'s) because they get used to index into the source text (which wants to be a `usize`), so using `.try_into().unwrap()` to convert `usize` -> `isize`

src/compiler/types.rs
```diff
@@ -1252,6 +1321,26 @@ pub enum Ternary {
     True = -1,
 }
 
+impl TryFrom<i32> for Ternary {
```
helixbass, line 1324: Per [this answer](https://stackoverflow.com/a/57578431/732366), convert `i32` -> enum variant (used below in the implementation of `BitAndAssign`)

src/compiler/types.rs
```diff
@@ -1438,43 +1531,100 @@ impl DiagnosticRelatedInformationInterface for Diagnostic {
...
+                diagnostic_with_detached_location.message_text()
+            }
+        }
+    }
 }
 
 impl DiagnosticInterface for Diagnostic {}
 
+#[derive(Clone, Debug)]
+pub enum DiagnosticMessageText {
```
helixbass, line 1550: This corresponds to `message_text: string | DiagnosticMessageChain` in the Typescript diagnostic types

src/compiler/checker.rs
peterstuart: 🤯

src/compiler/debug.rs
```diff
@@ -20,11 +20,21 @@ impl DebugType {
...
             self.fail(message);
         }
     }
+
+    pub fn check_defined<TValue>(&self, value: Option<TValue>, message: Option<&str>) -> TValue {
+        self.assert_is_defined(&value, message);
+        value.unwrap()
+    }
+
+    pub fn assert_never<TValue>(&self, value: TValue, message: Option<&str>) -> ! {
```
peterstuart, line 34: What motivated the `!` return type here?

src/compiler/emitter.rs
peterstuart: Why does this live outside the `Printer` impl?

src/compiler/factory/base_node_factory.rs
```diff
@@ -1,8 +1,68 @@
...
+}
+
+#[allow(non_snake_case)]
+impl BaseNodeFactory for BaseNodeFactoryConcrete {
+    fn create_base_source_file_node(&self, kind: SyntaxKind) -> BaseNode {
+        let mut SourceFileConstructor = self.SourceFileConstructor.borrow_mut();
+        if SourceFileConstructor.is_none() {
+            *SourceFileConstructor = Some(object_allocator.get_source_file_constructor());
+        }
+        (SourceFileConstructor.unwrap())(kind, -1, -1)
```
peterstuart, line 42: Combined with the if-statement above, this seems equivalent to `unwrap_or`?

src/compiler/factory/node_factory.rs
```diff
@@ -240,3 +273,57 @@ impl NodeFactory {
...
+// lazy_static! {
+//     pub static ref synthetic_factory: BaseNodeFactorySynthetic = BaseNodeFactorySynthetic::new();
+// }
+
+// pub fn get_synthetic_factory() -> &'static impl BaseNodeFactory {
+//     &*synthetic_factory
+// }
+
+pub fn get_synthetic_factory() -> BaseNodeFactorySynthetic {
+    BaseNodeFactorySynthetic::new()
```
peterstuart, line 294: This seems like it could live in the impl as well?

src/compiler/emitter.rs
helixbass: I guess the pattern I've been using for these Typescript helpers that return an object with a bunch of closure "methods" is to keep the calling interface the same (so just calling a simple `create_printer()` function) but then have the Rust version create an instance of a struct type under the hood (so just the difference of calling as `create_printer()` being more similar to the Typescript version than eg `Printer::create()`)

src/compiler/debug.rs
helixbass: In Typescript it's `never`, it'll always panic via `self.fail()` and so that helps it know about control flow at the point of calling `assert_never()`

src/compiler/factory/base_node_factory.rs
helixbass: Would that accomplish the mutation of `self.SourceFileConstructor`?

peterstuart: Oh I see. No if the mutation needs to persist beyond this function that doesn't work.

PR #17, opened 2021-12-25
Parse + bind array literal

helixbass:
In this PR:
- parse and bind an array literal eg `[1, 2, 3]`

To test:
If you run `cargo run tmp.tsx` against a source file containing eg `[1, 2, 3]` you should see debug-logging of a `SourceFile` with an `ArrayLiteralExpression` with three `elements` that are `NumericLiteral`'s with `text` `"1"`/`"2"`/`"3"`. Then in the `post-binding` debug-logging you should see that the `parent` links are initialized in those `NumericLiteral`'s. Then it should panic during type-checking

Based on `variable-declaration-type-checking-cont`


peterstuart (commented):
Nice, seems like you are far enough along now that parsing new things doesn't take much code.

PR #18, opened 2021-12-26
Parse + bind array literal declaration

helixbass:
In this PR:
- parse and bind an array literal declaration like `const a: number[] = [1, 2, true];`

I started working on the type-checking for this but ran into the fact that the definition of the "global" `Array<>` type (which is needed for resolving the type of `number[]` during type-checking) comes from one of the Typescript lib `.d.ts` files so decided to punt on that for now

To test:
If you run `cargo run tmp.tsx` against a source file containing eg `const a: number[] = [1, 2, true];` it should debug-log a `SourceFile` where the `type_` of the `VariableDeclaration` is an `ArrayTypeNode` whose `element_type` is a `NumberKeyword` `KeywordTypeNode`, should debug-log a `post-binding:` `SourceFile` with initialized `parent` references and then panic during type-checking

Based on `parse-array-literal`

### DIFF
--- a/src/compiler/binder.rs
+++ b/src/compiler/binder.rs
@@ -227,34 +227,26 @@ impl BinderType {
 
     fn bind_children(&self, node: &Node) {
         match node {
-            Node::Statement(statement) => match statement {
-                Statement::ExpressionStatement(expression_statement) => {
-                    self.bind_expression_statement(expression_statement);
-                }
-                _ => {
-                    self.bind_each_child(node);
-                }
-            },
-            Node::Expression(expression) => match expression {
-                Expression::PrefixUnaryExpression(_) => {
-                    self.bind_prefix_unary_expression_flow(node);
-                }
-                Expression::BinaryExpression(_) => unimplemented!(),
-                _ => {
-                    self.bind_each_child(node);
-                }
-            },
-            Node::SourceFile(source_file) => {
-                self.bind_each_functions_first(&source_file.statements);
+            Node::Statement(Statement::ExpressionStatement(expression_statement)) => {
+                self.bind_expression_statement(expression_statement);
             }
-            Node::VariableDeclarationList(_) => {
-                self.bind_each_child(node);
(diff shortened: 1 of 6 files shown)


src/compiler/binder.rs
```diff
@@ -227,34 +227,26 @@ impl BinderType {
...
-            Node::Expression(expression) => match expression {
-                Expression::PrefixUnaryExpression(_) => {
-                    self.bind_prefix_unary_expression_flow(node);
-                }
-                Expression::BinaryExpression(_) => unimplemented!(),
-                _ => {
-                    self.bind_each_child(node);
-                }
-            },
+            Node::Statement(Statement::ExpressionStatement(expression_statement)) => {
```
helixbass, line 230: Per a comment you'd made, I think "flattening" these node sub-variant `match` cases is a good idea for keeping the structure more similar to the Typescript version

In this case, that enabled just having a single "default" `self.bind_each_child(node);` at the end instead of having to nest it as a default inside each sub-variant

src/compiler/checker.rs
```diff
@@ -457,13 +548,27 @@ impl TypeChecker {
...
         self.get_conditional_flow_type_of_type(self.get_type_from_type_node_worker(node), node)
     }
 
     fn get_type_from_type_node_worker(&self, node: &Node /*TypeNode*/) -> Rc<Type> {
-        match node.kind() {
-            SyntaxKind::NumberKeyword => self.number_type(),
+        let node = match node {
+            Node::TypeNode(type_node) => type_node,
+            _ => panic!("Expected TypeNode"),
+        };
```
peterstuart, line 563: I bet you could make a `match_as!` macro to de-boilerplate this pattern a bit. (Probably a declarative macro would be fine for that?)

peterstuart (approved):

src/compiler/checker.rs
helixbass: Ya that would be a nice shorthand

PR #19, opened 2021-12-26
Parse + bind interface declaration

helixbass:
In this PR:
- parse and bind a simple interface declaration like:
```
interface Foo {
  bar: number
}
```

To test:
If you run `cargo run tmp.tsx` against a source file whose contents are:
```
interface Foo {
  bar: number
}
```
then you should see debug-logging of a `SourceFile` whose only statement is an `InterfaceDeclaration` with `name: "Foo"` and `members:` contains a `PropertySignature` with `name: "bar"` and `type_:` a `KeywordTypeNode` with `kind: NumberKeyword`. And in the `post-binding:` debug-logging you should see that the `locals` symbol table on the `SourceFile` contains a `Symbol` for `"Foo"`, which itself has a nested `members` symbol table containing a `Symbol` for `"bar"`. Then it should panic during type-checking

Based on `array-literal-declaration`

### DIFF
--- a/src/compiler/binder.rs
+++ b/src/compiler/binder.rs
@@ -10,7 +10,7 @@ use crate::{
     create_symbol_table, for_each, for_each_child, get_escaped_text_of_identifier_or_literal,
     get_name_of_declaration, is_binding_pattern, is_property_name_literal, object_allocator,
     set_parent, set_value_declaration, Expression, ExpressionStatement, NamedDeclarationInterface,
-    Node, NodeArray, NodeInterface, Statement, SymbolFlags,
+    Node, NodeArray, NodeInterface, PropertySignature, Statement, SymbolFlags, TypeElement,
 };
 
 bitflags! {
@@ -39,6 +39,7 @@ struct BinderType {
     file: RefCell<Option<Rc</*SourceFile*/ Node>>>,
     parent: RefCell<Option<Rc<Node>>>,
     container: RefCell<Option<Rc<Node>>>,
+    block_scope_container: RefCell<Option<Rc<Node>>>,
     Symbol: RefCell<Option<fn(SymbolFlags, __String) -> Symbol>>,
 }
 
@@ -47,6 +48,7 @@ fn create_binder() -> BinderType {
         file: RefCell::new(None),
         parent: RefCell::new(None),
         container: RefCell::new(None),
+        block_scope_container: RefCell::new(None),
         Symbol: RefCell::new(None),
     }
 }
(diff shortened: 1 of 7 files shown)


src/compiler/binder.rs
```diff
@@ -131,6 +153,18 @@ impl BinderType {
...
 
+        if symbol_flags.intersects(
+            SymbolFlags::Class
+                | SymbolFlags::Interface
+                | SymbolFlags::TypeLiteral
+                | SymbolFlags::ObjectLiteral,
+        ) {
+            let mut members = symbol.maybe_members();
+            if members.is_none() {
+                *members = Some(create_symbol_table());
```
helixbass, line 164: This pattern feels a little less clunky than in places where I've exposed a separate eg `.set_members()` setter - here, `.maybe_members()` returns a `RefMut<Option<SymbolTable>>`

src/compiler/scanner.rs
```diff
@@ -7,11 +7,16 @@ use std::iter::FromIterator;
...
+pub fn token_is_identifier_or_keyword(token: SyntaxKind) -> bool {
+    token >= SyntaxKind::Identifier
+}
+
 lazy_static! {
     static ref text_to_keyword_obj: HashMap<String, SyntaxKind> =
         HashMap::from_iter(IntoIter::new([
             ("const".to_string(), SyntaxKind::ConstKeyword),
             ("false".to_string(), SyntaxKind::FalseKeyword),
+            ("interface".to_string(), SyntaxKind::InterfaceKeyword),
```
peterstuart, line 19: It doesn't really matter, but I bet this could be a `HashMap<'static &str, SyntaxKind>` and you could avoid the conversions.

src/compiler/types.rs
```diff
@@ -713,6 +730,85 @@ pub struct ExpressionStatement {
...
+        self.type_.clone()
+    }
+
+    fn set_type(&mut self, type_: Rc<Node>) {
+        self.type_ = Some(type_);
+    }
+}
+
+#[derive(Debug)]
+#[ast_type(impl_from = false, interfaces = "NamedDeclarationInterface")]
```
peterstuart, line 766: Is it required that the things inside the parentheses be key-value pairs? If `impl_from` is only used when you set it to false, I think having an `omit_impl_from` flag might be a little more idiomatic (from my limited use of fancy macros).

src/compiler/scanner.rs
helixbass: I want to say I tried that and ran into something but can't really remember

src/compiler/types.rs
helixbass: Ya that makes sense, I'm not sure what the combination of `syn` + `darling` supports out-of-the-box

PR #20, opened 2021-12-26
Parse + bind object literal

helixbass:
In this PR:
- parse + bind an object literal

To test:
If you run `cargo run tmp.tsx` against a source file that contains:
```
interface Foo {
  bar: number
}

const a: Foo = {
  baz: 1,
}
```
then in the `post-binding:` debug-logging you should see in the `SourceFile` `statements` first an `InterfaceDeclaration` (per previous PR) and then a `VariableStatement`/`VariableDeclaration` where the `initializer` is an `ObjectLiteralExpression` whose `properties` contains a single `PropertyAssignment` with `name: "Baz"` and `initializer` a `NumericLiteral` with `text: "1"` and whose `type_` is a `TypeReferenceNode` with `type_name: "Foo"`

Based on `parse-interface-declaration`


src/compiler/types.rs
```diff
@@ -954,6 +1014,15 @@ impl Symbol {
     }
 }
 
+pub struct InternalSymbolName;
+
+#[allow(non_snake_case)]
+impl InternalSymbolName {
+    pub fn Object() -> __String {
+        __String::new("__object".to_string())
```
helixbass, line 1022: `InternalSymbolName` is a string-valued enum in the Typescript version (where the values are treated as the special `__String` type). I didn't see a way that the pattern of grabbing these known `__String`'s wouldn't currently involve copying them (since the things consuming this are expecting to own a `__String`) so just modeled this as getters that return a new `__String`

peterstuart (commented):
Looks good—I see the expected behavior.

PR #21, opened 2021-12-30
Type-check object literal declaration

helixbass:
In this PR:
- type-check basic object literal variable declaration against annotated interface type

To test:
```
interface Foo {
  bar: number
}

const a: Foo = {
  bar: true,
}
```
you should see a diagnostic with `message_text: "Type 'boolean' is not assignable to type 'number'."` debug-logged
```
interface Foo {
  bar: number
}

const a: Foo = {
  bar: 1,
}
```
you shouldn't see any diagnostics debug-logged


src/compiler/types.rs
```diff
@@ -1097,33 +1208,111 @@ impl Type {
...
             Type::IntrinsicType(intrinsic_type) => intrinsic_type.flags(),
             Type::LiteralType(literal_type) => literal_type.flags(),
+            Type::ObjectType(object_type) => object_type.flags(),
             Type::UnionOrIntersectionType(union_or_intersection_type) => {
                 union_or_intersection_type.flags()
             }
         }
     }
+
+    fn maybe_symbol(&self) -> Option<Rc<Symbol>> {
```
helixbass, line 1244: It's probably time to implement a similar boilerplate-generation macro for type types as I did for node types

src/compiler/checker.rs
```diff
@@ -1026,12 +2465,30 @@ impl NodeBuilder {
...
+        &self,
+        tracker: Option<&dyn SymbolTracker>,
+        cb: TCallback,
     ) -> Option<TReturn> {
-        let context = NodeBuilderContext::new(tracker.unwrap());
+        let default_tracker: Option<DefaultNodeBuilderContextSymbolTracker> = if tracker.is_some() {
+            None
+        } else {
+            Some(DefaultNodeBuilderContextSymbolTracker::new())
+        };
```
peterstuart, line 2489: A `match` seems like the more idiomatic thing to do here, to me.

peterstuart (commented):
This is more than I can really wrap my head around, so I skimmed some of it, but left a few comments.

src/compiler/checker.rs
helixbass: Ok updated to use `match`

src/compiler/binder.rs
```diff
@@ -162,7 +162,7 @@ impl BinderType {
         ) {
             let mut members = symbol.maybe_members();
             if members.is_none() {
-                *members = Some(create_symbol_table());
+                *members = Some(Rc::new(RefCell::new(create_symbol_table())));
```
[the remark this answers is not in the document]
helixbass: That's an interesting idea, but seems like then during creation (before I've wrapped in an `Rc`/`RefCell`) I'd need "temporary" "non-wrapped" versions of the types?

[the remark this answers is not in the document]
peterstuart: Right... I guess the macro could generate some private `BlahInternal` type based on the struct `Blah`? Seems pretty complicated.

PR #22, opened 2022-01-01
Flag unknown property on object literal

helixbass:
In this PR:
- flag a type error when an unknown property exists on an object literal that's initializing a variable of a certain interface type

To test:
```
interface Foo {
  bar: number
}

const a: Foo = {
  bar: 1,
  baz: 2,
}
```
then you should see a diagnostic debug-logged with chained `message_text` `"Type '{ baz: number; bar: number; }' is not assignable to type 'Foo'."` (not sure why `baz` is printed before `bar`, probably not sorting something correctly or something) + `"Object literal may only specify known properties, and 'baz' does not exist in type 'Foo'."`

Based on `type-check-object-literal`


src/compiler/checker.rs
```diff
@@ -611,11 +623,11 @@ impl TypeChecker {
...
 
-    fn type_to_string(&self, type_: &Type) -> String {
+    fn type_to_string(&self, type_: Rc<Type>) -> String {
         let writer = Rc::new(RefCell::new(create_text_writer("")));
-        let type_node = self
-            .node_builder
-            .type_to_type_node(type_, Some(&*(*writer).borrow()));
+        let type_node =
+            self.node_builder
+                .type_to_type_node(self, type_, Some(&*(*writer).borrow()));
```
helixbass, line 630: Most of the `NodeBuilder` methods need to get a `&TypeChecker` reference injected to be able to call back to type-checker methods

src/compiler/checker.rs
```diff
@@ -634,7 +646,11 @@ impl TypeChecker {
         result
     }
 
-    fn get_type_names_for_error_display(&self, left: &Type, right: &Type) -> (String, String) {
+    fn get_type_names_for_error_display(
+        &self,
+        left: Rc<Type>,
```
helixbass, line 651: Some method called from here needed to return an `Rc<Type>` so until I implement the same pattern (presumably with the help of a macro) for `Type`'s as I did for `Node`'s (so that I can eg call `.type_wrapper()` to get an `Rc<Type>` from a `&TypeInterface`) the `Rc`-ness needs to recursively "infect" the callers

src/compiler/checker.rs
```diff
@@ -2448,20 +2587,25 @@ fn create_node_builder() -> NodeBuilder {
...
-pub struct NodeBuilder {}
+pub struct NodeBuilder {
+    synthetic_factory: BaseNodeFactorySynthetic,
+}
 
 impl NodeBuilder {
     pub fn new() -> Self {
-        Self {}
+        Self {
+            synthetic_factory: get_synthetic_factory(),
```
helixbass, line 2597: Per a previous PR comment because of not being able to use `lazy_static!` for the `synthetic_factory` that the default singleton node `factory` wants to use as its `BaseNodeFactory` I'm instead calling `get_synthetic_factory()` to be able to inject that synthetic factory into calls into `factory` methods. So for `NodeBuilder` which is frequently calling into `factory` I figured I can give it a single synthetic factory instance upon creation and have it inject that rather than have the boilerplate + cost of calling `get_synthetic_factory()` inside any individual `NodeBuilder` method that needs to call into `factory`

src/compiler/checker.rs
```diff
@@ -2493,37 +2637,175 @@ impl NodeBuilder {
...
+    fn create_type_node_from_object_type(
+        &self,
+        type_checker: &TypeChecker,
+        context: &NodeBuilderContext,
+        type_: Rc<Type /*ObjectType*/>,
+    ) -> TypeNode {
+        let resolved = type_checker.resolve_structured_type_members(type_);
+
+        let members = self.create_type_nodes_from_resolved_type(type_checker, context, resolved);
+        let type_literal_node = factory.create_type_literal_node(&self.synthetic_factory, members);
```
helixbass, line 2737: A "type literal" is the anonymous type that is associated with an object literal, I guess

src/compiler/checker.rs
```diff
@@ -2787,19 +3160,38 @@ impl<'type_checker> CheckTypeRelatedTo<'type_checker> {
...
+        {
             return Ternary::True;
         }
 
+        let is_performing_excess_property_checks = !intersection_state
+            .intersects(IntersectionState::Target)
+            && (self.type_checker.is_object_literal_type(source.clone())
+                && get_object_flags(&*source).intersects(ObjectFlags::FreshLiteral));
+        if is_performing_excess_property_checks {
+            if self.has_excess_properties(source.clone(), target.clone(), report_errors) {
```
helixbass, line 3183: This is the entry point for actually checking for "excess properties" and then creating a diagnostic if it finds any

src/compiler/scanner.rs
```diff
@@ -63,7 +63,27 @@ pub fn skip_trivia(text: &str, pos: isize) -> isize {
         return pos;
     }
 
-    pos
+    let mut pos = pos as usize;
```
helixbass, line 66: I believe these are safe `usize`/`isize` conversions but would probably be better to use the `.try_from()` versions or whatever would panic if it can't actually correctly convert between the two

src/compiler/checker.rs
```diff
@@ -963,20 +995,34 @@ impl TypeChecker {
...
+        &self,
+        types: &[Rc<Type>],
+        exclude_kinds: TypeFlags,
+    ) -> ObjectFlags {
+        let mut result = ObjectFlags::None;
+        for type_ in types {
+            if !type_.flags().intersects(exclude_kinds) {
+                result |= get_object_flags(&*type_);
+            }
+        }
```
peterstuart, line 1022: This looks like it could be accomplished with a `fold`.

src/compiler/checker.rs
```diff
@@ -1148,20 +1220,31 @@ impl TypeChecker {
...
-        self.get_union_type_from_sorted_list(type_set)
+        let object_flags = (if includes.intersects(TypeFlags::NotPrimitiveUnion) {
+            ObjectFlags::None
+        } else {
+            ObjectFlags::PrimitiveUnion
+        }) | (if includes.intersects(TypeFlags::Intersection) {
+            ObjectFlags::ContainsIntersections
+        } else {
+            ObjectFlags::None
+        });
```
peterstuart, line 1239: I find the if-statements inside parentheses, or-ed together pretty hard to parse here.

src/compiler/checker.rs
```diff
@@ -2493,37 +2637,175 @@ impl NodeBuilder {
...
+        let object_flags = get_object_flags(&*type_);
+
+        if type_.flags().intersects(TypeFlags::TypeParameter)
+            || object_flags.intersects(ObjectFlags::ClassOrInterface)
+        {
+            return if let Some(symbol) = type_.maybe_symbol() {
+                self.symbol_to_type_node(type_checker, symbol, context, SymbolFlags::Type)
+            } else {
+                unimplemented!()
+            };
```
peterstuart, line 2684: Just personal preference, but I'd tend to only use `if let`s if I wasn't going to have an `else`, since with the `else` it's pretty much equivalent to a `match` with the final match being `_`.

src/compiler/checker.rs
```diff
@@ -2548,6 +2830,56 @@ impl NodeBuilder {
...
+        let symbol = chain[index].clone();
+
+        let mut symbol_name: Option<String>;
+        if index == 0 {
+            symbol_name =
+                Some(type_checker.get_name_of_symbol_as_written(symbol.clone(), Some(context)));
+        } else {
+            unimplemented!()
+        }
+        if symbol_name.is_none() {
```
peterstuart, line 2871: This seems impossible to be true, but is that just because the `None` would come from the thing that's currently `unimplemented`?

src/compiler/checker.rs
```diff
@@ -2787,19 +3160,38 @@ impl<'type_checker> CheckTypeRelatedTo<'type_checker> {
...
+            .intersects(IntersectionState::Target)
+            && (self.type_checker.is_object_literal_type(source.clone())
+                && get_object_flags(&*source).intersects(ObjectFlags::FreshLiteral));
+        if is_performing_excess_property_checks {
+            if self.has_excess_properties(source.clone(), target.clone(), report_errors) {
+                if report_errors {
+                    self.report_relation_error(
+                        head_message,
+                        source,
+                        if false { original_target } else { target },
```
peterstuart, line 3188: It'd be totally redundant since if-statements are expressions, but I really wish Rust had the `? :` ternary operator.

src/compiler/checker.rs
```diff
@@ -2838,6 +3230,80 @@ impl<'type_checker> CheckTypeRelatedTo<'type_checker> {
...
+        report_errors: bool,
+    ) -> bool {
+        if !self
+            .type_checker
+            .is_excess_property_check_target(target.clone())
+            || false
+        {
+            return false;
+        }
+        let is_comparing_jsx_attributes = false;
```
peterstuart, line 3246: I guess I'd never really thought about how TS is actually aware of JSX, rather than it being pre-processed into normal TS before it hits the compiler. (That actually seems kind of odd to me.)

peterstuart: Right, I didn’t even think about that aspect. So it must basically do the expansion to a function, type-check it, and then apply errors back to the JSX syntax?

PR #23, opened 2022-01-01
Split up `checker.rs` into multiple files

helixbass:
In this PR:
- split up `src/compiler/checker.rs` into multiple files organized by line numbers in corresponding Typescript `src/compiler/checker.ts`

To test:
Should still behave the same as of #22


src/compiler/checker/lines_0_1000.rs
```diff
@@ -0,0 +1,271 @@
+#![allow(non_upper_case_globals)]
+
+use bitflags::bitflags;
+use std::cell::{RefCell, RefMut};
+use std::collections::HashMap;
+use std::rc::Rc;
+
+use super::create_node_builder;
```
helixbass, line 8: For things that are only visible within `src/compiler/checker.ts` I used `super::` imports and `pub(super)` visibility

peterstuart: Nice, I've never seen `pub(super)` used.

PR #25, opened 2022-01-01
Parse + bind generic interface declaration

helixbass:
In this PR:
- parse and bind an interface declaration with a simple generic type parameter and a type annotation with a simple generic type argument

To test:
```
interface Foo<T> {
  bar: T
}

const a: Foo<number> = {
  bar: true,
}
```
then you should see `post-binding:` debug-logging where the `InterfaceDeclaration`'s `type_parameters` contains a `TypeParameterDeclaration` whose `name` is `"T"`, and then in the `type_` on the `VariableDeclaration` the `type_arguments` contains a `KeywordTypeNode` of `kind: NumberKeyword` (and these should have their `parent` references initialized)

Based on `multiple-checker-modules`


src/compiler/types.rs
```diff
@@ -890,22 +941,40 @@ impl PropertyAssignment {
     }
 }
 
+pub trait HasTypeParametersInterface {
```
helixbass, line 944: I don't think this corresponds to an `interface` in the Typescript version but it looks like there are a bunch of AST node types that have an optional `typeParameters?: NodeArray<TypeParameterDeclaration>` field so probably will make sense to have them all implement this shared interface for reading that field

src/compiler/types.rs
```diff
@@ -890,22 +941,40 @@ impl PropertyAssignment {
...
+    fn maybe_type_parameters(&self) -> Option<&NodeArray>;
```
peterstuart, line 945: You could give this a default implementation of `None` if there are types that will never have this populated.

peterstuart (commented):
Looks good, I see the expected output.

src/compiler/types.rs
helixbass: Ya that's a good thing to keep in mind, there's been one or two places where it seemed like maybe a default trait method implementation might be useful

PR #26, opened 2022-01-02
Generate diagnostic messages from JSON file

helixbass:
In this PR:
- mimic part of the `scripts/processDiagnosticMessages.js` in the Typescript version to generate a version of `src/compiler/diagnostic_information_map_generated.rs` that includes all diagnostic messages

To test:
If you `cargo install rust-script`, delete `src/compiler/diagnostic_information_map_generated.rs` and then run (from the repo root) `./scripts/process_diagnostic_messages.rs src/compiler/diagnosticMessages.json` it should re-generate `src/compiler/diagnostic_information_map_generated.rs`
Behavior should still be the same as of #25


scripts/process_diagnostic_messages.rs
```diff
@@ -0,0 +1,182 @@
+#!/usr/bin/env rust-script
```
helixbass, line 1: [`rust-script`](https://github.com/fornwall/rust-script) looked like a nice way for writing "scripts" in Rust

scripts/process_diagnostic_messages.rs
```diff
@@ -0,0 +1,182 @@
...
+    let input_file_path = Path::new(&args[1])/*.replace(/\\/g, "/")*/;
+    println!("Reading diagnostics from {}", input_file_path.display());
+    let input_str = fs::read_to_string(input_file_path).expect("Couldn't read input file");
+    let diagnostic_messages_json: HashMap<String, DiagnosticMessageSpec> =
+        serde_json::from_str(&input_str).expect("Couldn't parse input file as JSON");
+    let diagnostic_messages = diagnostic_messages_json;
+    let output_files_dir = input_file_path
+        .parent()
+        .expect("Couldn't get input file path parent");
+    let this_file_path_rel = pathdiff::diff_paths(
```
helixbass, line 47: Per [this answer](https://stackoverflow.com/a/46382269/732366), [`pathdiff`](https://docs.rs/pathdiff/latest/pathdiff/index.html) lets you get a relative path between two absolute paths (which mimics how the Typescript version was using NodeJS's `path.relative()`)

scripts/process_diagnostic_messages.rs
```diff
@@ -0,0 +1,182 @@
...
+enum DiagnosticCategory {
+    Error,
+    Message,
+    Suggestion,
+}
+
+#[derive(Clone, Copy, Debug, Deserialize)]
+struct DiagnosticMessageSpec {
+    category: DiagnosticCategory,
+    code: u32,
```
helixbass, line 28: It looks like there are some other optional fields on this but I haven't needed them yet so I can presumably just come back to this script and add them as `Option<...>` fields and print them as additional arguments to `diag()` when I need them

scripts/process_diagnostic_messages.rs
```diff
@@ -0,0 +1,182 @@
...
+        .map(|(key, value)| (key.to_string(), *value))
+        .collect();
+    entries.sort_by(|a, b| a.1.code.cmp(&b.1.code));
+    for (name, DiagnosticMessageSpec { code, category }) in entries.into_iter() {
+        let prop_name = convert_property_name(&name);
+        result.push_str(&format!(
+            "    pub const {}: DiagnosticMessage = diag(
+        {},
+        DiagnosticCategory::{:?},
+        {:?},
```
helixbass, line 132: This is printing strings with enclosing double quotes

src/compiler/diagnosticMessages.json
```diff
@@ -0,0 +1,7214 @@
+{
```
helixbass: This is the "source" `.json` file copied in from the Typescript version

scripts/process_diagnostic_messages.rs
```diff
@@ -0,0 +1,182 @@
...
+
+#[derive(Clone, Copy, Debug, Deserialize)]
+struct DiagnosticMessageSpec {
+    category: DiagnosticCategory,
+    code: u32,
+}
+
+fn main() {
+    let args = env::args().collect::<Vec<_>>();
+    if args.len() < 2 {
```
peterstuart, line 33: Maybe `!= 2`?

peterstuart (commented):
Looks good. Did you consider writing this as a macro? It might be impractically slow to be doing all this parsing/generating as part of every compile, but I think it'd feel a little cleaner.

scripts/process_diagnostic_messages.rs
helixbass: Ya that would make sense, the Typescript version uses `<` (it's actually checking `< 3` since it expects to be invoked like `node [script_name] [json_file_path]` - I think technically this could be a little fragile in that regard because you could invoke `rust-script` scripts either "directly" via the shebang like this or as `rust-script ...` (in which case the right number here would be `3`, not `2`))

helixbass (commented):
> Did you consider writing this as a macro?

Ya that crossed my mind, doing it this way is more similar to the Typescript version and ya does seem like it'd be unnecessary cost to regenerate on every compile since the diagnostics JSON should change infrequently

PR #27, opened 2022-01-06
Type-check generic interface variable declaration

helixbass:
In this PR:
- type-check a variable declaration for a generic interface + object literal initializer

To test:
```
interface Foo<T> {
  bar: T
}

const a: Foo<number> = {
  bar: true,
}
```
then you should see a diagnostic debug-logged with `message_text: "Type 'boolean' is not assignable to type 'number'."`

```
interface Foo<T> {
  bar: T
}

const a: Foo<number> = {
  bar: 1,
}
```
then you should see no diagnostics debug-logged


src/compiler/checker/lines_16000_18000.rs
```diff
@@ -132,6 +140,195 @@ impl TypeChecker {
...
+
+    pub(super) fn make_composite_type_mapper(
+        &self,
+        mapper1: TypeMapper,
+        mapper2: TypeMapper,
+    ) -> TypeMapper {
+        TypeMapper::new_composite(mapper1, mapper2)
+    }
+
+    pub(super) fn make_merged_type_mapper(
```
helixbass, line 219: This isn't a separate function in the Typescript version, there `makeCompositeTypeMapper()` takes a `kind` argument for the `kind` "discriminant" field on the `TypeMapper` type, but since I'm modeling `TypeMapper` as an enum (without a `kind` field) I just made separate `make_composite_type_mapper()`/`make_merged_type_mapper()` functions that create different enum variants

src/compiler/checker/lines_18000_20000.rs
```diff
@@ -261,7 +262,7 @@ impl<'type_checker> CheckTypeRelatedTo<'type_checker> {
             if self.should_check_as_excess_property(prop.clone(), source.symbol()) && true {
                 if !self.type_checker.is_known_property(
                     reduced_target.clone(),
-                    &prop.escaped_name,
+                    prop.escaped_name(),
```
helixbass, line 265: Since `Symbol` became an enum it now uses `SymbolInterface` methods everywhere to access/modify "fields"

src/compiler/checker/lines_32000_40000.rs
```diff
@@ -179,7 +183,26 @@ impl TypeChecker {
...
+                Some(node.as_has_type_arguments().maybe_type_arguments().unwrap()),
+                |type_argument, _| self.get_type_from_type_node(type_argument),
+            ),
+            Some(type_parameters),
+        )
+        .unwrap()
+    }
+
     pub(super) fn check_type_reference_node(&mut self, node: &TypeReferenceNode) {
+        maybe_for_each(node.type_arguments.as_ref(), |type_argument, _| {
```
helixbass, line 204: I guess I added a `maybe_for_each()` variant of `for_each()` whose argument is an `Option`

src/compiler/checker/lines_8000_12000.rs
```diff
@@ -144,7 +153,24 @@ impl TypeChecker {
         type_
     }
 
+    pub(super) fn get_type_of_instantiated_symbol(&self, symbol: &Symbol) -> Rc<Type> {
+        let links = self.get_symbol_links(symbol);
+        let mut links = links.borrow_mut();
+        if links.type_.is_none() {
+            let type_ = self.instantiate_type(
```
helixbass, line 160: This is actually what will return a `number` type by looking up the `T` type parameter in the `TypeMapper` that maps `T` -> `number`

src/compiler/checker/lines_8000_12000.rs
```diff
@@ -161,19 +187,147 @@ impl TypeChecker {
...
+                type_parameters,
+                self.get_declared_type_of_type_parameter(
+                    self.get_symbol_of_node(&**declaration).unwrap(),
+                ),
+            ));
+        }
+        type_parameters
+    }
+
+    pub(super) fn get_outer_type_parameters(
```
helixbass, line 206: I think that "outer" type parameters might be like "closed-over" generic parameters from enclosing declarations?

src/compiler/types.rs
```diff
@@ -1071,11 +1112,13 @@ pub trait TypeCheckerHost: ModuleSpecifierResolutionHost {
...
 }
 
+#[derive(Debug)]
 #[allow(non_snake_case)]
 pub struct TypeChecker {
     pub _types_needing_strong_references: RefCell<Vec<Rc<Type>>>,
-    pub Symbol: fn(SymbolFlags, __String) -> Symbol,
+    pub Symbol: fn(SymbolFlags, __String) -> BaseSymbol,
     pub Type: fn(TypeFlags) -> BaseType,
+    pub(crate) empty_symbols: Rc<RefCell<SymbolTable>>,
```
helixbass, line 1121: This isn't different than the others, I think `pub(crate)` would probably make more sense for all of these `TypeChecker` members (they just need to be accessible from the type-checking code)

src/compiler/types.rs
```diff
@@ -1157,60 +1202,455 @@ bitflags! {
...
+}
+
+pub trait TransientSymbolInterface: SymbolInterface {
+    fn symbol_links(&self) -> Rc<RefCell<SymbolLinks>>;
+    fn check_flags(&self) -> CheckFlags;
+}
+
+#[derive(Debug)]
+pub enum TransientSymbol {
+    BaseTransientSymbol(BaseTransientSymbol),
```
helixbass, line 1442: It looks like there will be other sub-variants of `TransientSymbol` so made it an enum

src/compiler/types.rs
```diff
@@ -2411,6 +2912,100 @@ impl From<BaseInterfaceType> for InterfaceType {
     }
 }
 
+pub trait InterfaceTypeWithDeclaredMembersInterface {
```
helixbass, line 2915: I think similar to `ResolvedType` `InterfaceTypeWithDeclaredMembers` is something that `InterfaceType`'s can "mutate into" at some point in their lifecycle, so again seems like could maybe use a typestate pattern or something but for now just always have `InterfaceType` implement `InterfaceTypeWithDeclaredMembers`

src/compiler/checker/lines_0_1000.rs
```diff
@@ -9,9 +9,24 @@ use super::create_node_builder;
 use crate::{
     __String, create_diagnostic_collection, create_symbol_table, object_allocator,
     DiagnosticCollection, FreshableIntrinsicType, IntrinsicType, Number, ObjectFlags, Symbol,
-    SymbolFlags, SymbolTable, Type, TypeChecker, TypeCheckerHost, TypeFlags,
+    SymbolFlags, SymbolId, SymbolInterface, SymbolTable, Type, TypeChecker, TypeCheckerHost,
+    TypeFlags,
 };
 
+thread_local! {
```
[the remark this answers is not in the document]
peterstuart: Is that a built-in macro?

src/compiler/checker/lines_16000_18000.rs
```diff
@@ -132,6 +140,195 @@ impl TypeChecker {
...
+        targets: Option<Vec<Rc<Type>>>,
+    ) -> TypeMapper {
+        if sources.len() == 1 {
+            self.make_unary_type_mapper(
+                sources[0].clone(),
+                if let Some(targets) = targets {
+                    targets[0].clone()
+                } else {
+                    self.any_type()
+                },
```
peterstuart, line 155: I'd usually use `map_or_unwrap` for things like this.

src/compiler/checker/lines_16000_18000.rs
```diff
@@ -132,6 +140,195 @@ impl TypeChecker {
...
+    ) -> TypeMapper {
+        TypeMapper::new_merged(mapper1, mapper2)
+    }
+
+    pub(super) fn combine_type_mappers(
+        &self,
+        mapper1: Option<TypeMapper>,
+        mapper2: TypeMapper,
+    ) -> TypeMapper {
+        if let Some(mapper1) = mapper1 {
```
peterstuart, line 232: Same comment as above about `map_or_unwrap`.

src/compiler/checker/lines_16000_18000.rs
```diff
@@ -132,6 +140,195 @@ impl TypeChecker {
...
+        mut symbol: Rc<Symbol>,
+        mapper: &TypeMapper,
+    ) -> Rc<Symbol> {
+        let links = self.get_symbol_links(&symbol);
+        let links = links.borrow();
+        if let Some(type_) = links.type_.as_ref() {
+            if !self.could_contain_type_variables(type_.clone()) {
+                return symbol;
+            }
+        }
```
peterstuart, line 262: I don't know that it's an improvement, but looking for built-in ways to accomplish this same thing got me here:

```rs
let links = links.borrow();
if links
    .type_
    .as_ref()
    .filter(|type_| !self.could_contain_type_variables((*type_).clone()))
    .is_some()
{
    return symbol;
}
```

Generally I find if-statements about `Option`s/`Result`s to be kind of a smell, because there are so many convenience functions built in.

src/compiler/checker/lines_16000_18000.rs
```diff
@@ -132,6 +140,195 @@ impl TypeChecker {
...
+        &self,
+        type_: Option<Rc<Type>>,
+        mapper: Option<&TypeMapper>,
+    ) -> Option<Rc<Type>> {
+        if let Some(type_) = type_.as_ref() {
+            if let Some(mapper) = mapper {
+                return Some(self.instantiate_type_with_alias(type_.clone(), mapper, None, None));
+            }
+        }
+        type_
```
peterstuart, line 303: Unless you're trying to mimic a larger if-statement, I'd write this with a match:

```rs
match (type_.as_ref(), mapper) {
    (Some(type_), Some(mapper)) => {
        return Some(self.instantiate_type_with_alias(type_.clone(), mapper, None, None))
    }
    _ => type_,
}
```

src/compiler/core.rs
```diff
@@ -35,6 +51,25 @@ pub fn every<TItem, TCallback: FnMut(&TItem, usize) -> bool>(
...
+    let mut result: Option<Vec<_>> = None;
+    if let Some(array) = array {
+        let mut some_result = vec![];
+        for (i, item) in array.into_iter().enumerate() {
+            some_result.push(f(item, i));
+        }
+        result = Some(some_result);
+    }
+    result
+}
```
peterstuart, line 71: I'd probably write this function like this:

```rs
array.map(|array| {
    array
        .into_iter()
        .enumerate()
        .map(|(i, item)| f(item, i))
        .collect()
})
```

src/compiler/utilities.rs
```diff
@@ -10,13 +10,30 @@ use crate::{
...
+    if let Some(declarations) = declarations {
+        for declaration in declarations {
+            if declaration.kind() == kind {
+                return Some(declaration.clone());
+            }
+        }
+    }
+
+    None
+}
```
peterstuart, line 35: I think this is equivalent:

```rs
symbol
    .maybe_declarations()
    .as_ref()
    .and_then(|declarations| {
        declarations.iter().find_map(|declaration| {
            if declaration.kind() == kind {
                Some(declaration.clone())
            } else {
                None
            }
        })
    })
```

src/compiler/checker/lines_0_1000.rs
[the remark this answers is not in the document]
helixbass: Yup it's `std::thread_local`

src/compiler/checker/lines_16000_18000.rs
helixbass: Ah I think I'm used to reaching for `.unwrap_or()`/`.unwrap_or_else()` but not `.map_or()`/`.map_or_else()`, updated to use `.map_or_else()`

src/compiler/checker/lines_16000_18000.rs
helixbass: In this case it's unhappy that I'm trying to "own" `mapper2` twice if I use `.map_or()`/`.map_or_else()` since I guess it's not smart enough to know that eg in `.map_or_else()` only one of the two closures will be executed

src/compiler/checker/lines_16000_18000.rs
helixbass: Ah ya that's pretty nice. The thing that would read closest to the Typescript version would be:
```
if let Some(type_) = links.type_.as_ref() && !self.could_contain_type_variables(type_.clone()) {
  ...
}
```
but I think having "hybrid" `if let`/"normal" `if` conditions like that is only supported on nightly or something?

src/compiler/checker/lines_16000_18000.rs
helixbass: Ok ya this isn't straying too far from the Typescript version (which uses a ternary `type && mapper ? ... : ...`) as a way of expressing conditional "if this/these are defined, do this, otherwise do this" I guess, updated to use `match`

src/compiler/checker/lines_16000_18000.rs
peterstuart: Yeah I think so

PR #28, opened 2022-01-07
Parse + bind string literal

helixbass:
In this PR:
- parse + bind a string literal type + string literal value

To test:
If you run `cargo run tmp.tsx` against a source file which contains eg `const a: 'foo' = 'baz'`, you should see `post-binding:` debug-logging where the `VariableDeclaration`'s `type_` is a `LiteralTypeNode` whose `literal` is a `StringLiteral` with `text: "foo"` and the `VariableDeclaration`'s `initializer` is a `StringLiteral` with `text: "baz"` (and both should have `parent` links initialized)

Based on `type-check-generic-interface`


src/compiler/checker/lines_4000_8000.rs
```diff
@@ -292,13 +292,12 @@ impl NodeBuilder {
...
                 .create_literal_type_node(
                     &self.synthetic_factory,
-                    &*Into::<Rc<Node>>::into(
-                        if type_.as_intrinsic_type().intrinsic_name() == "true" {
-                            factory.create_true(&self.synthetic_factory)
-                        } else {
-                            factory.create_false(&self.synthetic_factory)
-                        },
-                    ),
+                    if type_.as_intrinsic_type().intrinsic_name() == "true" {
```
helixbass, line 295: I'd previously commented that here was a place where I had to make sure to wrap the node in an `Rc<Node>` before passing along, I think it makes more sense for the `.create_literal_type_node()` method to just take an `Rc<Node>` since that's what it's storing and then there's no way to forget to do that

src/compiler/parser.rs
```diff
@@ -386,7 +397,11 @@ impl ParserType {
         self.set_parse_diagnostics(vec![]);
         self.set_parsing_context(ParsingContext::None);
 
-        self.scanner_mut().set_text(Some(_source_text), None, None);
+        let mut scanner = self.scanner_mut();
+        scanner.set_text(Some(_source_text), None, None);
+        // scanner.set_on_error(Some(Box::new(move |message, length| {
```
helixbass, line 402: The scanner in the Typescript version has an optional `onError` callback field which the parser initializes to call back into its `scanError()` function. I tried messing around with actually mimicking that (so storing a closure in the `Scanner` that captures a reference to the `ParserType` instance) but was running into lifetime stuff and I think conceptually it wouldn't make sense because if the scanner was effectively holding onto an immutable reference to the parser that should mean that the parser could never be used mutably while the scanner existed, which we don't want. So instead I used the same technique you suggested previously elsewhere of "injecting" the `on_error` callback when calling into scanner methods that might need to invoke it (so far just the one entry point from the parser of `scanner.scan()`)

src/compiler/parser.rs
```diff
@@ -462,7 +477,9 @@ impl ParserType {
     }
 
     fn next_token_without_check(&self) -> SyntaxKind {
-        let current_token = self.scanner().scan();
+        let current_token = self
+            .scanner()
+            .scan(Some(&|message, length| self.scan_error(message, length)));
```
helixbass, line 482: Per above comment, this is where instead I'm injecting the `on_error` callback into the scanner when calling `.scan()`

src/compiler/factory/node_factory.rs
```diff
@@ -108,24 +108,41 @@ impl NodeFactory {
...
-            text: value.to_string(),
-        }
+        BaseLiteralLikeNode::new(node, value)
     }
 
     pub fn create_numeric_literal<TBaseNodeFactory: BaseNodeFactory>(
         &self,
         base_factory: &TBaseNodeFactory,
-        value: &str,
+        value: String,
```
peterstuart, line 120: I'd generally push the conversion of `&str` into `String` until the last possible place, but maybe that doesn't make sense here.

src/compiler/types.rs
```diff
@@ -3339,6 +3402,9 @@ impl CharacterCodes {
     pub const max_ascii_character: char = '';
 
     pub const line_feed: char = '\n';
+    pub const carriage_return: char = '\r';
+    pub const line_separator: char = ' ';
+    pub const paragraph_separator: char = ' ';
```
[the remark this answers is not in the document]
peterstuart: I didn't even know line/paragraph separators were characters that existed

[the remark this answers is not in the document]
helixbass: I think they're "Unicode characters" (not in the ASCII range)

PR #29, opened 2022-01-07
Type-check string literal

helixbass:
In this PR:
- type-check a variable declaration of a string literal type

To test:
If you run `cargo run tmp.tsx` against a source file containing eg `const a: 'foo' = 'baz'` you should see a diagnostic debug-logged with message `Type '"baz"' is not assignable to type '"foo"'.`
If you run `cargo run tmp.tsx` against a source file containing eg `const a: 'foo' = 'foo'` you shouldn't see any diagnostics debug-logged

Based on `parse-string-literal`


src/compiler/checker/lines_0_1000.rs
```diff
@@ -27,6 +27,20 @@ pub(super) fn increment_next_symbol_id() {
     });
 }
 
+thread_local! {
```
helixbass, line 30: This mimics the pattern with `SymbolId`'s/`SymbolLinks`

src/compiler/checker/lines_4000_8000.rs
```diff
@@ -246,24 +324,28 @@ impl NodeBuilder {
...
         cb: TCallback,
     ) -> Option<TReturn> {
         let default_tracker: Option<DefaultNodeBuilderContextSymbolTracker> = match tracker {
             Some(_) => None,
             None => Some(DefaultNodeBuilderContextSymbolTracker::new()),
         };
-        let context =
-            NodeBuilderContext::new(tracker.unwrap_or_else(|| default_tracker.as_ref().unwrap()));
+        let context = NodeBuilderContext::new(
+            flags.unwrap_or(NodeBuilderFlags::None),
```
peterstuart, line 346: You might prefer the explicitness you have here, but `unwrap_or_default` could be nice if `NoneBuilderFlags` implemented `Default`.

src/compiler/utilities.rs
```diff
@@ -397,6 +470,15 @@ impl DiagnosticCollection {
     }
 }
 
+fn escape_non_ascii_string(
+    s: &str,
+    quote_char: Option<
+        char, /*CharacterCodes.doubleQuote | CharacterCodes.singleQuote | CharacterCodes.backtick*/
+    >,
+) -> String {
+    s.to_string()
```
peterstuart, line 479: Is this something that'll be expanded later?

PR #30, opened 2022-01-08
Parse + bind union of string literals type

helixbass:
In this PR:
- parse + bind a union-of-string-literals type

To test:
If you run `cargo run tmp.tsx` against a source file containing eg `const a: 'foo' | 'bar' = 'baz'`, you should see `post-binding:` debug-logging where the `VariableDeclaration`'s `type_` is a `UnionTypeNode` whose `types` are a `LiteralTypeNode` with `StringLiteral` `"foo"` and a `LiteralTypeNode` with `StringLiteral` `"bar"` (and they should have their `parent` bindings initialized)

Based on `type-check-string-literal`

### DIFF
--- a/src/compiler/factory/node_factory.rs
+++ b/src/compiler/factory/node_factory.rs
@@ -1,21 +1,25 @@
+#![allow(non_upper_case_globals)]
+
 use std::rc::Rc;
 
 use crate::{
-    create_base_node_factory, escape_leading_underscores, ArrayLiteralExpression, ArrayTypeNode,
-    BaseBindingLikeDeclaration, BaseGenericNamedDeclaration, BaseInterfaceOrClassLikeDeclaration,
-    BaseLiteralLikeNode, BaseNamedDeclaration, BaseNode, BaseNodeFactory, BaseNodeFactoryConcrete,
-    BaseVariableLikeDeclaration, BinaryExpression, EmptyStatement, Expression, ExpressionStatement,
-    Identifier, InterfaceDeclaration, LiteralLikeNodeInterface, LiteralTypeNode, Node, NodeArray,
-    NodeArrayOrVec, NodeFactory, NodeFlags, NodeInterface, NumericLiteral, ObjectLiteralExpression,
+    create_base_node_factory, escape_leading_underscores, is_omitted_expression, last_or_undefined,
+    ArrayLiteralExpression, ArrayTypeNode, BaseBindingLikeDeclaration, BaseGenericNamedDeclaration,
+    BaseInterfaceOrClassLikeDeclaration, BaseLiteralLikeNode, BaseNamedDeclaration, BaseNode,
+    BaseNodeFactory, BaseNodeFactoryConcrete, BaseVariableLikeDeclaration, BinaryExpression,
+    EmptyStatement, Expression, ExpressionStatement, Identifier, InterfaceDeclaration,
+    IntersectionTypeNode, LiteralLikeNodeInterface, LiteralTypeNode, Node, NodeArray,
+    NodeArrayOrVec, NodeFactory, NodeFlags, NumericLiteral, ObjectLiteralExpression,
     PrefixUnaryExpression, PropertyAssignment, PropertySignature, SourceFile, StringLiteral,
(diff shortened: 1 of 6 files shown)


src/compiler/parser.rs
```diff
@@ -1171,19 +1194,67 @@ impl ParserType {
...
+        let is_union_type = operator == SyntaxKind::BarToken;
+        let has_leading_operator = self.parse_optional(operator);
+        let mut type_: Option<TypeNode> = Some(if has_leading_operator {
+            self.parse_function_or_constructor_type_to_error(is_union_type)
+                .unwrap_or_else(|| parse_constituent_type(self))
+        } else {
+            parse_constituent_type(self)
+        });
+        if self.token() == operator || has_leading_operator {
+            let mut types: Vec<Node> = vec![type_.take().unwrap().into()];
```
helixbass, line 1220: Here I gave `type_` an `Option` wrapper so that I could move it here (by "taking" it) but still reassign to it later (to be able to return it at the end of the function)

src/compiler/parser.rs
```diff
@@ -83,6 +83,12 @@ pub fn for_each_child<TNodeCallback: FnMut(Option<Rc<Node>>), TNodesCallback: Fn
         Node::TypeNode(TypeNode::ArrayTypeNode(array_type)) => {
             visit_node(&mut cb_node, Some(array_type.element_type.clone()))
         }
+        Node::TypeNode(TypeNode::UnionTypeNode(union_type)) => {
+            visit_nodes(&mut cb_node, &mut cb_nodes, Some(&union_type.types))
```
helixbass, line 87: These are a single case in the Typescript version where it casts the node to `UnionOrIntersectionTypeNode` (aka `UnionTypeNode | IntersectionTypeNode`), I could achieve something similar here by creating eg a `UnionOrIntersectionTypeNodeInterface` trait and having `UnionTypeNode` and `IntersectionTypeNode` implement it but seemed like possibly overkill (if other places keep wanting to treat a `UnionTypeNode`/`IntersectionTypeNode` symmetrically then I'll probably go ahead and do that)

src/compiler/factory/node_factory.rs
```diff
@@ -298,7 +336,24 @@ impl NodeFactory {
...
+                last_or_undefined(Into::<&[Rc<Node>]>::into(&elements_as_node_array));
+            if let Some(last_element) = last_element {
+                if is_omitted_expression(&**last_element) {
+                    Some(true)
+                } else {
+                    None
+                }
+            } else {
+                None
+            }
```
peterstuart, line 354: I'd write this as an `and_then` on `last_element`.

src/compiler/parser.rs
```diff
@@ -801,15 +820,18 @@ impl ParserType {
         let mut list: Vec<Node> = vec![];
         let list_pos = self.get_node_pos();
 
+        let mut comma_start: isize = -1;
```
peterstuart, line 823: You can also do this in a way where you avoid the type annotation on the variable:

```rs
let mut comma_start = -1_isize;
```

I'm not really sure which I prefer 🤷🏻

src/compiler/types.rs
```diff
@@ -441,6 +444,12 @@ impl From<&NodeArray> for Vec<Rc<Node>> {
     }
 }
 
+impl<'node_array> From<&'node_array NodeArray> for &'node_array [Rc<Node>] {
```
peterstuart, line 447: Does Rust force you to provide the lifetime annotation here? (I'm a little surprised it doesn't infer it, but I've never tried to implement a trait for a slice.)

src/compiler/factory/node_factory.rs
helixbass: Sure, updated to use `.and_then()`

src/compiler/parser.rs
helixbass: Ah ok ya that numeric-type-suffix reads a little strangely to me

src/compiler/types.rs
helixbass: Ya it does require it, it seems to not infer that the lifetime of the `&NodeArray` in the type parameter `From<&NodeArray>` is obviously the same as the lifetime of the corresponding `&NodeArray` in the method argument `node_array: &NodeArray`

PR #31, opened 2022-01-08
Type-check union of string literals

helixbass:
In this PR:
- type-check variable declaration of union-of-string-literals type

To test:
If you run `cargo run tmp.tsx` against a source file containing eg `const a: 'foo' | 'bar' = 'baz'` you should see a diagnostic debug-logged with message `Type '"baz"' is not assignable to type '"foo" | "bar"'.`
If you run `cargo run tmp.tsx` against a source file containing eg `const a: 'foo' | 'bar' = 'bar'` you shouldn't see any diagnostics debug-logged

Based on `parse-union-of-string-literals`


src/compiler/checker/lines_18000_20000.rs
```diff
@@ -321,8 +321,17 @@ impl<'type_checker> CheckTypeRelatedTo<'type_checker> {
         &self,
         source: Rc<Type>,
         target: &UnionOrIntersectionType,
+        report_errors: bool,
     ) -> Ternary {
         let target_types = target.types();
+        if target.flags().intersects(TypeFlags::Union) {
+            if self
+                .type_checker
+                .contains_type(target_types, source.clone())
```
helixbass, line 330: I guess the simple case for a "yes is assignable" to a union-type target is when the source "is" one of the union'ed target types (based on a type ID comparison), which is the case here since string literal types are cached to be unique

src/compiler/checker/lines_32000_40000.rs
```diff
@@ -216,6 +216,20 @@ impl TypeChecker {
         self.check_source_element(Some(&*node.element_type));
     }
 
+    pub(super) fn check_union_or_intersection_type(
+        &mut self,
+        node: &Node, /*UnionOrIntersectionTypeNode*/
+    ) {
+        for_each(
+            node.as_union_or_intersection_type_node().types(),
```
helixbass, line 224: Per previous PR comment, I ran into a couple other places where it wanted to treat union + intersection types symmetrically (to grab their `types` field) so went ahead and created a `UnionOrIntersectionTypeNodeInterface` + `Node.as_union_or_intersection_type_node()` "unwrapper helper"

src/compiler/emitter.rs
```diff
@@ -209,6 +212,14 @@ impl Printer {
         self.write_punctuation("}");
     }
 
+    fn emit_union_type(&mut self, node: &UnionTypeNode) {
+        self.emit_list(
+            Some(node),
+            Some(&node.types),
+            ListFormat::UnionTypeConstituents,
```
helixbass, line 219: This is sort of cool how a bitflags "compound value" is like a set of configuration options, in this case `ListFormat::UnionTypeConstituents` includes `ListFormat::BarDelimited` so when this `ListFormat` gets passed down to `.write_delimiter()` that will see that flag set and write a `|`

src/compiler/checker/lines_12000_16000.rs
```diff
@@ -368,6 +368,17 @@ impl TypeChecker {
         unimplemented!()
     }
 
+    pub(super) fn get_type_id(&self, type_: Rc<Type>) -> TypeId {
+        type_.id()
+    }
+
+    pub(super) fn contains_type(&self, types: &[Rc<Type>], type_: Rc<Type>) -> bool {
+        /*binary_search(...)*/
```
peterstuart, line 376: Is the list (eventually) sorted in some way that makes a binary search possible?

src/compiler/checker/lines_4000_8000.rs
```diff
@@ -542,6 +572,35 @@ impl NodeBuilder {
...
+                    .flags
+                    .intersects(NodeBuilderFlags::UseFullyQualifiedType);
+                let mut result: Vec<Rc<Node>> = vec![];
+                let mut i: usize = 0;
+                for type_ in types {
+                    i += 1;
+                    let type_node =
+                        self.type_to_type_node_helper(type_checker, type_.clone(), context);
+                    result.push(type_node.into());
+                }
```
peterstuart, line 596: Unless you're going to do something exciting with `i`, maybe you could iterate over the `.enumerate` of `types` instead of manually managing the index?

src/compiler/checker/lines_12000_16000.rs
helixbass: Ya I guess it's not codified at the type level (eg by using `SortedArray`) but it does appear that the `.types` field on union/intersection types will always be sorted by type ID (and I'm mimicking that now in `.add_type_to_union()`)

So updated here to use `binary_search_copy_key()` now that that exists

PR #32, opened 2022-01-08
Type-check variable usage

helixbass:
In this PR:
- type-check usage of a variable

To test:
```
const a = true
++a
```
then you should see a diagnostic debug-logged with message `"An arithmetic operand must be of type 'any', 'number', 'bigint' or an enum type."`
```
const a = 1
++a
```
then you should see no diagnostics debug-logged

Based on `type-check-union-of-string-literals`

### DIFF
--- a/src/compiler/checker/lines_8000_12000.rs
+++ b/src/compiler/checker/lines_8000_12000.rs
@@ -166,10 +166,13 @@ impl TypeChecker {
             type_ = self
                 .try_get_type_from_effective_type_node(&*declaration)
                 .unwrap_or_else(|| {
-                    self.check_property_assignment(match &*declaration {
-                        Node::PropertyAssignment(property_assignment) => property_assignment,
-                        _ => panic!("Expected PropertyAssignment"),
-                    })
+                    self.check_property_assignment(
+                        match &*declaration {
+                            Node::PropertyAssignment(property_assignment) => property_assignment,
+                            _ => panic!("Expected PropertyAssignment"),
+                        },
+                        None,
+                    )
                 });
         } else if is_property_signature(&*declaration) || is_variable_declaration(&*declaration) {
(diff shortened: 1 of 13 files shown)


src/compiler/checker/lines_8000_12000.rs
```diff
@@ -88,17 +88,27 @@ impl TypeChecker {
...
         }
-        unimplemented!()
+
+        if has_only_expression_initializer(declaration)
+            && declaration
+                .as_has_expression_initializer()
+                .maybe_initializer()
+                .is_some()
+        {
+            let type_ = self.check_declaration_initializer(declaration, None);
```
helixbass, line 98: Here's where we're inferring the type of a variable declaration based on the initializer when it didn't have an explicit type annotation

src/compiler/checker/lines_8000_12000.rs
```diff
@@ -88,17 +88,27 @@ impl TypeChecker {
...
     pub(super) fn get_widened_type_for_variable_like_declaration(
         &self,
         declaration: &Node,
     ) -> Rc<Type> {
-        self.widen_type_for_variable_like_declaration(
-            self.get_type_for_variable_like_declaration(declaration),
-            declaration,
-        )
+        let type_ = self.get_type_for_variable_like_declaration(declaration);
+        self.widen_type_for_variable_like_declaration(type_, declaration)
```
helixbass, line 111: I think this change ended up being unnecessary, at one point I was trying to change some `TypeChecker` methods to take `&mut self` but unsurprisingly that "bubbled up" to somewhere where that wasn't going to work (eg where a closure that captures a mutable `self` reference gets passed to another `TypeChecker` method) so backed it out (I don't think I had to then eg wrap any `TypeChecker` fields in a `RefCell` so maybe there were just "dangling" existing `&mut self`'s that didn't need to be `&mut self`'s)

peterstuart (approved):

PR #33, opened 2022-01-09
"Type type" macro

helixbass:
In this PR:
- use a new `#[type_type]` procedural macro for `Type` enum variants similar to how `#[ast_type]` macro is used for `Node` variants

Not in this PR:
- updating places that wrap `Type`'s in `Rc`'s to use the generated `.into()` (which encapsulates setting the `_type_wrapper` field on the `BaseType`) and changing function parameters to prefer being `&Type` instead of `Rc<Type>`

To test:

Based on `type-check-variable-usage`


local-macros/src/lib.rs
```diff
@@ -50,7 +50,7 @@ impl AstTypeArgs {
     }
 }
 
-fn get_struct_interface_impl(
+fn get_ast_struct_interface_impl(
```
helixbass, line 53: I tried moving the `ast_type` procedural macro implementation into a nested module but it told me that the `#[proc_macro_attribute]` attribute can only be used in the crate root

peterstuart: Huh, that's a weird restriction.

local-macros/src/lib.rs
```diff
@@ -550,3 +550,481 @@ pub fn ast_type(attr: TokenStream, item: TokenStream) -> TokenStream {
     }
     .into()
 }
+
+#[derive(Debug, FromMeta)]
```
peterstuart, line 554: `FromMeta` parses the args in `()`, right?

peterstuart (commented):
Nice, that's a hefty cleanup.

PR #34, opened 2022-01-09
Pass type references as function parameters

helixbass:
In this PR:
- now that `.type_wrapper()` exists, update function parameters to be `&Type` instead of `Rc<Type>`

To test:

Based on `type-type-macro`

### DIFF
--- a/local-macros/src/lib.rs
+++ b/local-macros/src/lib.rs
@@ -658,9 +658,8 @@ fn get_type_struct_interface_impl(
                     fn get_or_initialize_fresh_type(
                         &self,
                         type_checker: &crate::TypeChecker,
-                        wrapper: &::std::rc::Rc<crate::Type>,
                     ) -> Rc<Type> {
-                        self.#first_field_name.get_or_initialize_fresh_type(type_checker, wrapper)
+                        self.#first_field_name.get_or_initialize_fresh_type(type_checker)
                     }
 
                     fn regular_type(&self) -> ::std::rc::Rc<crate::Type> {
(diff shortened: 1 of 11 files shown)


local-macros/src/lib.rs
```diff
@@ -658,9 +658,8 @@ fn get_type_struct_interface_impl(
                     fn get_or_initialize_fresh_type(
                         &self,
                         type_checker: &crate::TypeChecker,
-                        wrapper: &::std::rc::Rc<crate::Type>,
```
helixbass, line 661: This explicit `wrapper` argument to `get_or_initialize_fresh_type()` is not longer necessary because we can call `self.type_wrapper()` to get the same thing

src/compiler/checker/lines_0_1000.rs
```diff
@@ -160,40 +160,36 @@ pub fn create_type_checker<TTypeCheckerHost: TypeCheckerHost>(
             )
             .into(),
     ));
-    type_checker.any_type = Some(Rc::new(
+    type_checker.any_type = Some(
```
helixbass, line 163: Like with AST node creation, always use `.into()` (which encapsulates initializing the `_type_wrapper` reference on `BaseType`) when wrapping in an `Rc<Type>`

src/compiler/checker/lines_0_1000.rs
```diff
@@ -160,40 +160,36 @@ pub fn create_type_checker<TTypeCheckerHost: TypeCheckerHost>(
...
+    )
+    .into();
     match &*true_type {
         Type::IntrinsicType(intrinsic_type) => match intrinsic_type {
             IntrinsicType::FreshableIntrinsicType(freshable_intrinsic_type) => {
                 freshable_intrinsic_type
                     .regular_type
-                    .init(&regular_true_type, true);
-                freshable_intrinsic_type.fresh_type.init(&true_type, true);
+                    .init(&regular_true_type, false);
```
helixbass, line 191: It's now not true that `regular_true_type` is a unique `Rc` reference because we've stored a second reference in `_type_wrapper`

src/compiler/types.rs
```diff
@@ -2265,8 +2253,35 @@ impl StringLiteralType {
     }
 }
 
+impl LiteralTypeInterface for StringLiteralType {
```
helixbass, line 2256: This was a mistake in the previous PR, this isn't actually a "boilerplate" trait implementation because `.get_or_initialize_fresh_type()` has a custom implementation

PR #35, opened 2022-01-09
symbol_type macro

helixbass:
In this PR:
- use a new `#[symbol_type]` procedural macro for `Symbol` enum variants similar to how `#[ast_type]` macro is used for `Node` variants

Not in this PR:
- updating places that wrap `Symbol`'s in `Rc`'s to use the generated `.into()` (which encapsulates setting the `_symbol_wrapper` field on the `BaseSymbol`) and changing function parameters to prefer being `&Symbol` instead of `Rc<Symbol>`

To test:

Based on `pass-type-references`

### DIFF
--- a/local-macros/src/lib.rs
+++ b/local-macros/src/lib.rs
@@ -1026,3 +1026,336 @@ pub fn type_type(attr: TokenStream, item: TokenStream) -> TokenStream {
     }
     .into()
 }
+
+#[derive(Debug, FromMeta)]
+struct SymbolTypeArgs {
+    #[darling(default)]
+    ancestors: Option<String>,
+    #[darling(default)]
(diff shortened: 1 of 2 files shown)


local-macros/src/lib.rs
```diff
@@ -1026,3 +1026,366 @@ pub fn type_type(attr: TokenStream, item: TokenStream) -> TokenStream {
     }
     .into()
 }
+
+#[derive(Debug, FromMeta)]
+struct SymbolTypeArgs {
+    #[darling(default)]
```
peterstuart, line 1032: What does the `default` do here? Default to `None`?

helixbass: I think so

PR #37, opened 2022-01-10
Type error for unknown variable reference

helixbass:
In this PR:
- fix bug where type-checking of non-generic interface variable declarations had stopped working
- flag type error when an unknown variable is referenced

To test:
The testing from #21/#22 should now work again
If you run `cargo run tmp.tsx` against a source file containing eg `++b` you should see a diagnostic debug-logged with message `Cannot find name 'b'.`. The testing from #32 should still work (ie it should still be able to look up an existing variable reference)

Based on `pass-symbol-references`

### DIFF
--- a/src/compiler/checker/lines_24000_32000.rs
+++ b/src/compiler/checker/lines_24000_32000.rs
@@ -20,6 +20,9 @@ impl TypeChecker {
         check_mode: Option<CheckMode>,
     ) -> Rc<Type> {
         let symbol = self.get_resolved_symbol(node);
+        if Rc::ptr_eq(&symbol, &self.unknown_symbol()) {
+            return self.error_type();
+        }
 
         let local_or_export_symbol = self
             .get_export_symbol_of_value_symbol_if_exported(Some(symbol))
--- a/src/compiler/checker/lines_8000_12000.rs
+++ b/src/compiler/checker/lines_8000_12000.rs
@@ -310,11 +310,13 @@ impl TypeChecker {
                 self.get_outer_type_parameters_of_class_or_interface(symbol);
             let local_type_parameters =
                 self.get_local_type_parameters_of_class_or_interface_or_type_alias(symbol);
+            let mut need_to_set_constraint = false;
             let type_: InterfaceType = if outer_type_parameters.is_some()
                 || local_type_parameters.is_some()
                 || kind == ObjectFlags::Class
                 || false
             {
+                need_to_set_constraint = true;
                 type_.set_object_flags(type_.object_flags() | ObjectFlags::Reference);
                 let mut this_type = self.create_type_parameter(Some(symbol.symbol_wrapper()));
                 this_type.is_this_type = Some(true);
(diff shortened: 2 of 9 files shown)


src/compiler/checker/lines_24000_32000.rs
```diff
@@ -20,6 +20,9 @@ impl TypeChecker {
         check_mode: Option<CheckMode>,
     ) -> Rc<Type> {
         let symbol = self.get_resolved_symbol(node);
+        if Rc::ptr_eq(&symbol, &self.unknown_symbol()) {
+            return self.error_type();
```
helixbass, line 24: First, when it tries to resolve the `Symbol` of an unknown variable reference that returns `unknownSymbol` (a singleton `Symbol` on the type checker instance), and then here when it's trying to get the type of the symbol it returns the singleton `Type` `errorType` from the type checker instance for an unknown symbol

`errorType` has the `TypeFlags.Any` flag set on it, and so that basically makes there be no further type errors raised from usage of the unknown variable (because `any` basically disables type-checking)

src/compiler/checker/lines_8000_12000.rs
```diff
@@ -333,16 +335,18 @@ impl TypeChecker {
...
-                Type::ObjectType(ObjectType::InterfaceType(InterfaceType::BaseInterfaceType(
-                    base_interface_type,
-                ))) => match &**base_interface_type.this_type.borrow_mut().as_ref().unwrap() {
-                    Type::TypeParameter(type_parameter) => {
-                        *type_parameter.constraint.borrow_mut() = Some(Rc::downgrade(&type_rc));
-                    }
-                    _ => panic!("Expected TypeParameter"),
-                },
-                _ => panic!("Expected BaseInterfaceType"),
+            if need_to_set_constraint {
```
helixbass, line 338: This is where the bug was that was causing the previously working type-checking of non-generic interface types to stop working, this setting of the `constraint` that had to happen after the the type was wrapped in an `Rc<Type>` (because it needed to get set to a weak reference to that `Rc<Type>`) should only have been happening in the "generic" case (corresponding to the `if` branch above)

src/compiler/checker/lines_8000_12000.rs
```diff
@@ -333,16 +335,18 @@ impl TypeChecker {
...
+                match &*type_rc {
+                    Type::ObjectType(ObjectType::InterfaceType(
+                        InterfaceType::BaseInterfaceType(base_interface_type),
+                    )) => match &**base_interface_type.this_type.borrow_mut().as_ref().unwrap() {
```
peterstuart, line 342: This is an intense line!

src/compiler/utilities.rs
```diff
@@ -761,8 +763,15 @@ fn create_file_diagnostic(
...
+    args: Option<Vec<String>>,
 ) -> DiagnosticWithLocation {
-    let text = get_locale_specific_message(message);
+    let mut text = get_locale_specific_message(message);
+
+    if let Some(args) = args {
+        if !args.is_empty() {
+            text = format_string_from_args(&text, args);
+        }
+    }
```
peterstuart, line 774: Just playing a little type tetris here, this also works to get rid of the nested if-statements (but is worse in other ways):

```rs
let args: Vec<_> = args.into_iter().flatten().collect();

if !args.is_empty() {
    text = format_string_from_args(&text, args);
}
```

src/compiler/checker/lines_8000_12000.rs
helixbass: It looks better as of #42

PR #38, opened 2022-01-10
Parse + bind if-statement

helixbass:
In this PR:
- parse + bind an `if`/`else` statement

To test:
```
if (true) {
    const a = 1;
} else {
    const b = 2;
}

++a
```
then you should see `post-binding:` debug-logging where the `SourceFile`'s first `Statement` is an `IfStatement` whose `expression` is a `TrueKeyword` and whose `then_statement` and `else_statement` are both `Block`'s whose `statements` contain a single `VariableStatement`, and `parent` bindings should be initialized in them. Then it should panic during type-checking

Based on `fix-object-literal-checking`


src/compiler/parser.rs
```diff
@@ -856,20 +885,26 @@ impl ParserType {
         self.create_node_array(list, list_pos, None, Some(comma_start >= 0))
     }
 
+    fn create_missing_list(&self) -> NodeArray {
+        let mut list = self.create_node_array(vec![], self.get_node_pos(), None, None);
+        list.is_missing_list = true;
```
helixbass, line 890: The Typescript version models `MissingList` as a sub-type of `NodeArray`, but that seems like overkill/a lot of wiring-up (to turn `NodeArray` into an enum) so I just made all `NodeArray`'s have an `is_missing_list: bool` field

src/compiler/types.rs
```diff
@@ -1059,6 +1086,24 @@ pub struct EmptyStatement {
     pub _node: BaseNode,
 }
 
+#[derive(Debug)]
+#[ast_type(ancestors = "Statement")]
+pub struct Block {
+    _node: BaseNode,
+    pub statements: NodeArray, /*<Statement>*/
+    pub(crate) multi_line: Option<bool>,
```
helixbass, line 1094: Some Typescript fields are marked `@internal` which I'm not sure exactly what it means (guessing it suppresses generating documentation for things?) but so `pub(crate)` seemed like possibly the appropriate corresponding "visibility"

src/compiler/binder.rs
```diff
@@ -295,6 +307,24 @@ impl BinderType {
...
+        action(self, value);
+    }
+
+    fn bind_condition<TNodeRef: Borrow<Node>>(&self, node: Option<TNodeRef>) {
+        self.do_with_conditional_branches(BinderType::bind, node);
+    }
+
+    fn bind_if_statement(&self, node: &IfStatement) {
+        self.bind_condition(Some(node.expression.clone()));
+        self.bind(Some(&*node.then_statement));
```
peterstuart, line 324: Could you explain what's going on here? (Is `then_statement` deref?)

src/compiler/factory/node_factory.rs
```diff
@@ -431,10 +443,28 @@ impl NodeFactory {
...
+        then_statement: Statement,
+        else_statement: Option<Statement>,
+    ) -> IfStatement {
+        let node = self.create_base_node(base_factory, SyntaxKind::IfStatement);
+        let node = IfStatement::new(
+            node,
+            expression.into(),
+            self.as_embedded_statement(Some(then_statement.into()))
+                .unwrap(),
+            self.as_embedded_statement(else_statement.map(|else_statement| else_statement.into())),
```
peterstuart, line 465: This would also work:

```rs
self.as_embedded_statement(else_statement.map(Into::into)),
```

PR #39, opened 2022-01-11
Type-check if-statement

helixbass:
In this PR:
- type-check `if`/`else` statement, flagging attempt to use a variable outside of its declared scope

To test:
```
if (true) {
    const a = 1;
} else {
    const b = 2;
}

++a
```
then you should see a diagnostic debug-logged with message `Cannot find name 'a'.`

```
if (true) {
    const a = 1;
    ++a
} else {
    const b = 2;
}
```
then you shouldn't see any diagnostics debug-logged

Based on `parse-if-statement`

### DIFF
--- a/src/compiler/checker/lines_1000_2000.rs
+++ b/src/compiler/checker/lines_1000_2000.rs
@@ -167,10 +167,20 @@ impl TypeChecker {
         let error_location = location.clone();
 
         while let Some(location_unwrapped) = location {
-            if location_unwrapped.maybe_locals().is_some()
-                && !self.is_global_source_file(&*location_unwrapped)
-            {
-                unimplemented!()
+            let location_maybe_locals = location_unwrapped.maybe_locals();
+            if let Some(location_locals) = &*location_maybe_locals {
+                if !self.is_global_source_file(&*location_unwrapped) {
+                    result = lookup(self, location_locals, name, meaning);
+                    if let Some(result_unwrapped) = result.as_ref() {
+                        let mut use_result = true;
+
+                        if use_result {
+                            break;
+                        } else {
+                            result = None;
+                        }
+                    }
+                }
             }
 
             match location_unwrapped.kind() {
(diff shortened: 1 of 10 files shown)


src/compiler/checker/lines_1000_2000.rs
```diff
@@ -167,10 +167,20 @@ impl TypeChecker {
...
 
         while let Some(location_unwrapped) = location {
-            if location_unwrapped.maybe_locals().is_some()
-                && !self.is_global_source_file(&*location_unwrapped)
-            {
-                unimplemented!()
+            let location_maybe_locals = location_unwrapped.maybe_locals();
+            if let Some(location_locals) = &*location_maybe_locals {
+                if !self.is_global_source_file(&*location_unwrapped) {
+                    result = lookup(self, location_locals, name, meaning);
```
helixbass, line 173: This is where it's attempting to resolve a name to a local variable

src/compiler/factory/node_factory.rs
```diff
@@ -493,7 +493,9 @@ impl NodeFactory {
         declarations: TDeclarations,
         flags: Option<NodeFlags>,
     ) -> VariableDeclarationList {
-        let node = self.create_base_node(base_factory, SyntaxKind::VariableDeclarationList);
+        let flags = flags.unwrap_or(NodeFlags::None);
+        let mut node = self.create_base_node(base_factory, SyntaxKind::VariableDeclarationList);
+        node.flags |= flags & NodeFlags::BlockScoped;
```
helixbass, line 498: This is what tags a variable declaration list as eg `NodeFlags::Const` (so that then later when doing name resolution it can treat `const`/`let` declarations differently from `var`, specifically I think `var` is function-scoped rather than block-scoped)

src/compiler/checker/lines_32000_40000.rs
```diff
@@ -318,6 +319,20 @@ impl TypeChecker {
         self.get_type_from_type_node(node);
     }
 
+    pub(super) fn check_block(&mut self, node: &Block) {
+        if is_function_or_module_block(node) {
+            for_each(&node.statements, |statement, _| {
+                self.check_source_element(Some(statement.clone()));
+                Option::<()>::None
```
peterstuart, line 326: What's this for?

helixbass: The `Option` return type of the callback is generic so it complains if it can't resolve it to a specific type

PR #40, opened 2022-01-12
Type-check bigint, heterogeneous union type

helixbass:
In this PR:
- parse + type-check bigint literals + type annotations
- type-check a "heterogeneous union" type

To test:
```
const a: bigint = 1n;

++a
```
then you shouldn't see any diagnostics debug-logged
```
const a: bigint = 1;

++a
```
then you should see a diagnostic debug-logged with message `Type 'number' is not assignable to type 'bigint'.`
```
const a: number = 1n;

++a
```
then you should see a diagnostic debug-logged with message `Type 'bigint' is not assignable to type 'number'.`
```
const a: number | bigint = 1;

++a
```
then you shouldn't see any diagnostics debug-logged
```
const a: boolean | number = 1;

++a
```
then you should see a diagnostic debug-logged with message `An arithmetic operand must be of type 'any', 'number', 'bigint' or an enum type.`

Based on `type-check-if-statement`


src/compiler/checker/lines_12000_16000.rs
```diff
@@ -390,7 +400,21 @@ impl TypeChecker {
             if false {
                 unimplemented!()
             } else {
-                type_set.push(type_.type_wrapper());
+                let len = type_set.len();
+                let index: isize = if len > 0 && type_.id() > type_set[len - 1].id() {
+                    !isize::try_from(len).unwrap()
+                } else {
+                    binary_search_copy_key(
```
helixbass, line 407: The `binary_search()`/`insert_sorted()`/`binary_search_key()` functions were where I was running into lifetime issues - I didn't really end up using the `for<'a> ...` (higher-kinded lifetimes?) technique from [this answer](https://stackoverflow.com/a/47126516/732366) but did sort of do what it's saying as far as "hardcoding" the fact that the "key selector" closure returns a reference in `binary_search()`/`binary_search_key()`, so exposed parallel helpers `binary_search_copy_key()`/`binary_search_key_copy_key()` for when the key selector returns a `Copy`-able value rather than a reference. I could maybe revisit this since I think it's possible I don't even need to hardcode the reference-returning-ness in `binary_search()`/`binary_search_key()`

So here since the key for binary search is a `TypeId` (which is `Copy`), I'm using `binary_search_copy_key()`

src/compiler/checker/lines_12000_16000.rs
```diff
@@ -390,7 +400,21 @@ impl TypeChecker {
...
+                        type_set,
+                        &type_.type_wrapper(),
+                        |type_, _| Some(self.get_type_id(type_)),
+                        compare_values,
+                        None,
+                    )
+                };
+                if index < 0 {
+                    type_set.insert(usize::try_from(!index).unwrap(), type_.type_wrapper());
```
helixbass, line 416: There's a lot of weird conversions between `isize` and `usize` because it wants to return negative values to mean something that should then be bit-negated (`~` in most languages, `!` in Rust) but then that value gets used as an index (which has to be a `usize` afaik)

src/compiler/core.rs
```diff
@@ -85,12 +86,27 @@ pub fn concatenate<TItem>(mut array1: Vec<TItem>, mut array2: Vec<TItem>) -> Vec
...
+fn identity_key_selector<TItem>(item: TItem, _: Option<usize>) -> TItem {
+    item
+}
+
+pub fn insert_sorted<TItem /*, TComparer: Comparer<&'array_or_item TItem>*/>(
+    array: &mut SortedArray<TItem>,
+    insert: TItem,
+    // compare: Comparer<&'array_or_item TItem>,
+    // compare: Comparer<&TItem, &TItem>,
+    compare: fn(&TItem, &TItem) -> Comparison,
```
helixbass, line 98: Per Slack, this is where I "inlined" the type alias `Comparer` since then it didn't seem to care about the lifetime of these `&TItem`'s

src/compiler/core.rs
```diff
@@ -85,12 +86,27 @@ pub fn concatenate<TItem>(mut array1: Vec<TItem>, mut array2: Vec<TItem>) -> Vec
     array1
 }
 
-pub fn insert_sorted<TItem>(array: &mut SortedArray<TItem>, insert: TItem) {
+fn identity_key_selector<TItem>(item: TItem, _: Option<usize>) -> TItem {
```
helixbass, line 89: I can delete this, ended up not using it (was when I was seeing if using `fn`'s instead of `Fn*`'s (ie closures) would get rid of any lifetime issues)

src/compiler/core_public.rs
```diff
@@ -28,3 +34,26 @@ impl<TItem: Clone> From<&SortedArray<TItem>> for Vec<TItem> {
         sorted_array._vec.clone()
     }
 }
+
+impl<TItem> Deref for SortedArray<TItem> {
```
helixbass, line 38: Allow a `SortedArray` to deref into a slice (`&[TItem]`)

src/compiler/types.rs
```diff
@@ -2714,31 +2816,57 @@ pub struct DiagnosticMessage {
...
         Self {
             _diagnostic_related_information: diagnostic_related_information,
+            related_information,
         }
     }
 }
 
 impl DiagnosticRelatedInformationInterface for BaseDiagnostic {
+    fn maybe_as_diagnostic(&self) -> Option<&Diagnostic> {
+        panic!("Shouldn't call maybe_as_diagnostic() on BaseDiagnostic")
```
helixbass, line 2863: This is kind of messy, in the Typescript version `DiagnosticRelatedInformation` is the base interface for `Diagnostic`, before I was just exposing a `DiagnosticRelatedInformationInterface` that `Diagnostic` implemented but now it looks like there are concrete `DiagnosticRelatedInformation`'s getting stored/passed around so to avoid `dyn DiagnosticRelatedInformationInterface`'s (that then I wouldn't be able to extract the sub-variant of if needed) I went ahead and made `DiagnosticRelatedInformation` a concrete "parent" enum of `Diagnostic`. But it seems like most places are just using `Diagnostic`'s so I don't really want to always wrap in/pass around the parent `DiagnosticRelatedInformation`. So then what I'm doing now is making things generic on `<TDiagnosticRelatedInformation: DiagnosticRelatedInformationInterface>` and then if I need to do something specific to `Diagnostic`'s rather than being able to `match` on the parent `DiagnosticRelatedInformation` I exposed a `maybe_as_diagnostic()` method on `DiagnosticRelatedInformationInterface` that returns `Some(diagnostic)` for `Diagnostic` and `None` for the other variant of `DiagnosticRelatedInformation` (`BaseDiagnosticRelatedInformation`). But the intent is only to call `maybe_as_diagnostic()` on these "top-level" implementors of `DiagnosticRelatedInformationInterface` so I'm panicking if it gets called on anything besides those. I guess another alternative would be to make diagnostics use the same pattern as `Node`/`Type` and always wrap them and have a `.diagnostic_wrapper()` method that returns the parent `DiagnosticRelatedInformation` (from which I could then determine if it's a `Diagnostic` or not by `match`'ing)

src/compiler/utilities.rs
```diff
@@ -476,24 +493,17 @@ impl DiagnosticCollection {
...
-        //     });
         if let Some(diagnostics) = self
             .file_diagnostics
             .get_mut(&diagnostic.file().unwrap().file_name)
         {
-            insert_sorted(diagnostics, diagnostic);
+            insert_sorted(
+                diagnostics,
+                diagnostic,
+                |rc_diagnostic_a: &Rc<Diagnostic>, rc_diagnostic_b: &Rc<Diagnostic>| {
```
helixbass, line 503: I guess naming this `rc_*` became a little unnecessary after it complained that it needed explicit type annotations here

src/compiler/checker/lines_12000_16000.rs
peterstuart: What does the bit-negation do? (Why would you negate in index?)

src/compiler/core_public.rs
```diff
@@ -28,3 +34,26 @@ impl<TItem: Clone> From<&SortedArray<TItem>> for Vec<TItem> {
...
+    type Target = [TItem];
```
peterstuart, line 39: Ah interesting. I guess I've never really seen the non-reference-type slice type anywhere before. That's not something that can actually exist, right? (An "owned" slice?)

src/compiler/types.rs
```diff
@@ -2756,7 +2884,28 @@ impl DiagnosticRelatedInformationInterface for BaseDiagnostic {
     }
 }
 
+impl DiagnosticInterface for BaseDiagnostic {
+    fn maybe_related_information(&self) -> Option<&[DiagnosticRelatedInformation]> {
+        self.related_information.as_deref()
```
peterstuart, line 2889: What does `as_deref` do (vs I guess `*`)?

helixbass: Here are the [docs](https://doc.rust-lang.org/std/option/enum.Option.html#method.as_deref), I guess it's like `.as_ref()` (on an `Option`) but also deref's the now-referenced content

PR #41, opened 2022-01-13
enum_unwrapped!() macro

helixbass:
In this PR:
- use new `enum_unwrapped!()` macro for boilerplate of unwrapping expected enum variants

To test:
Behavior should still be the same as of #40 

Based on `heterogeneous-union-type`

### DIFF
--- a/local-macros/src/lib.rs
+++ b/local-macros/src/lib.rs
@@ -1394,7 +1394,26 @@ pub fn symbol_type(attr: TokenStream, item: TokenStream) -> TokenStream {
     .into()
 }
 
-fn get_enum_unwrapped_match(argument: &Expr, ancestors: &[Ident]) -> TokenStream2 {}
+fn get_enum_unwrapped_match(argument: &Expr, ancestors: &[Ident]) -> TokenStream2 {
+    let mut unwrapped_variant_selector = quote! {
+        unwrapped
+    };
+    for (index, ancestor) in ancestors.iter().skip(1).rev().enumerate() {
+        let previous_ancestor = &ancestors[ancestors.len() - 2 - index];
+        unwrapped_variant_selector = quote! {
+            crate::#previous_ancestor::#ancestor(#unwrapped_variant_selector)
+        }
+    }
+
+    let last_variant = ancestors.last().unwrap();
+
+    quote! {
+        match #argument {
+            #unwrapped_variant_selector => unwrapped,
+            _ => panic!("Expected {}", stringify!(#last_variant)),
+        }
+    }
+}
 
 fn expr_to_ident(expr: &Expr) -> Result<Ident> {
     match expr {
@@ -1439,7 +1458,7 @@ pub fn enum_unwrapped(input: TokenStream) -> TokenStream {
         ancestors,
     } = parse_macro_input!(input as EnumUnwrapped);
 
-    let mut enum_match = get_enum_unwrapped_match(&argument, &ancestors);
+    let enum_match = get_enum_unwrapped_match(&argument, &ancestors);
 
     quote! {
         (#enum_match)
--- a/local-macros/src/lib.rs
+++ b/local-macros/src/lib.rs
@@ -1461,7 +1461,7 @@ pub fn enum_unwrapped(input: TokenStream) -> TokenStream {
     let enum_match = get_enum_unwrapped_match(&argument, &ancestors);
 
     quote! {
-        (#enum_match)
+        #enum_match
     }
     .into()
 }
--- a/local-macros/src/lib.rs
+++ b/local-macros/src/lib.rs
@@ -2,8 +2,12 @@ use darling::FromMeta;
 use proc_macro::TokenStream;
 use proc_macro2::{Ident, TokenStream as TokenStream2};
 use quote::quote;
+use syn::parse::{Parse, ParseStream, Result};
 use syn::Data::{Enum, Struct};
-use syn::{parse_macro_input, AttributeArgs, DataEnum, DeriveInput, Fields, FieldsNamed};
+use syn::{
+    parse_macro_input, AttributeArgs, DataEnum, DeriveInput, Error, Expr, ExprArray, Fields,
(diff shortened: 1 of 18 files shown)


local-macros/src/lib.rs
```diff
@@ -1389,3 +1393,75 @@ pub fn symbol_type(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+            crate::#previous_ancestor::#ancestor(#unwrapped_variant_selector)
+        }
+    }
+
+    let last_variant = ancestors.last().unwrap();
+
+    quote! {
+        match #argument {
+            #unwrapped_variant_selector => unwrapped,
+            _ => panic!("Expected {}", stringify!(#last_variant)),
```
helixbass, line 1413: I don't think I tested that this actually prints out the expected panic message (eg "Expected Expression")

local-macros/src/lib.rs
```diff
@@ -1389,3 +1393,75 @@ pub fn symbol_type(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+        }
+    }
+}
+
+fn expr_to_ident(expr: &Expr) -> Result<Ident> {
+    match expr {
+        Expr::Path(path) => match path.path.get_ident() {
```
helixbass, line 1420: For parsing the "array" of enum variant names (eg `[Node, Expression]`) `syn`'s `ExprArray` gives you a `Vec<Expr>`, and the enum variant of `Expr` corresponding to simple `Ident`'s is `Expr::Path` (of which an identifier is the simple case vs a path involving `::`). Then it gives you a convenience method `.get_ident()` for "unwrapping" an `Ident`

local-macros/src/lib.rs
```diff
@@ -1389,3 +1393,75 @@ pub fn symbol_type(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+
+impl Parse for EnumUnwrapped {
+    fn parse(input: ParseStream) -> Result<Self> {
+        let argument: Expr = input.parse()?;
+        input.parse::<Token![,]>()?;
+        let ancestors_expr: ExprArray = input.parse()?;
+        // let ancestors = ancestors
+        //     .elems
+        //     .iter()
+        //     .map(|expr| expr_to_ident(expr)?)
```
helixbass, line 1441: It seemed unhappy about something with trying to use `?` inside this `.map()` closure, I'm guessing there's some idiomatic way to do this?

local-macros/src/lib.rs
```diff
@@ -1389,3 +1393,75 @@ pub fn symbol_type(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+
+struct EnumUnwrapped {
+    argument: Expr,
+    ancestors: Vec<Ident>,
+}
+
+impl Parse for EnumUnwrapped {
+    fn parse(input: ParseStream) -> Result<Self> {
+        let argument: Expr = input.parse()?;
+        input.parse::<Token![,]>()?;
```
helixbass, line 1436: I was basing this off of an [example](https://github.com/dtolnay/syn/blob/master/examples/lazy-static/lazy-static/src/lib.rs) from `syn`, since it currently expects a known # of "arguments" this step-by-step parsing is pretty straightforward

src/compiler/binder.rs
```diff
@@ -420,12 +421,8 @@ impl BinderType {
         name: __String,
     ) -> Rc<Symbol> {
         let symbol = self.create_symbol(symbol_flags, name).wrap();
-        match &*self.file() {
-            Node::SourceFile(source_file) => {
-                source_file.keep_strong_reference_to_symbol(symbol.clone());
-            }
-            _ => panic!("Expected SourceFile"),
-        }
+        enum_unwrapped!(&*self.file(), [Node, SourceFile])
```
helixbass, line 424: The next step of convenience that's tempting is to give the macros knowledge of the `Node`/`Type`/etc enum variants so that I could just write eg:
```
node_unwrapped!(&*self.file(), SourceFile)
```

src/compiler/checker/lines_32000_40000.rs
```diff
@@ -193,10 +188,7 @@ impl TypeChecker {
         check_mode: Option<CheckMode>,
     ) -> Rc<Type> {
         self.check_expression_for_mutable_location(
-            match &*node.initializer {
-                Node::Expression(expression) => expression,
-                _ => panic!("Expected Expression"),
-            },
+            enum_unwrapped!(&*node.initializer, [Node, Expression]),
```
helixbass, line 191: Another possible convenience for unwrapping is where eg a `Node` variant has an `Rc<Node>` field that's of a known sub-variant (eg here we know that `PropertyAssignment.initializer` is always an `Expression`), expose a custom eg `.initializer_as_expression()` helper that does the unwrapping for you (although would lifetimes get weird if that tried to return a reference when called against an `Rc<Node>` variable?)

src/compiler/parser.rs
```diff
@@ -139,7 +140,7 @@ pub fn create_source_file(file_name: &str, source_text: &str) -> SourceFile {
     Parser().parse_source_file(file_name, source_text)
 }
 
-enum MissingNode {
+pub enum MissingNode {
```
helixbass, line 143: Here the macro's assumption that its top-level enum is in `crate::` scope was being violated

src/compiler/parser.rs
```diff
@@ -747,14 +748,14 @@ impl ParserType {
...
-        match self.create_missing_node(
-            SyntaxKind::Identifier,
-            diagnostic_message.unwrap_or(default_message),
-            Some(vec![msg_arg]),
-        ) {
-            MissingNode::Identifier(identifier) => identifier,
-            _ => panic!("Expected identifier"),
-        }
+        enum_unwrapped!(
+            self.create_missing_node(
```
helixbass, line 752: It would be cool if this could be written in more of a "pipeline" order (eg `self.create_missing_node(...).unwrapped!(Identifier)`) but I assume you can't write a "macro method" like that?

PR #42, opened 2022-01-13
.as_*() unwrapping helpers

helixbass:
In this PR:
- expose `.as_*()` "unwrapping helpers" on `Node`/`Type` to improve the ergonomics on top of `enum_unwrapped!()`

To test:
Behavior should still be the same as of #40 

Based on `enum-unwrapped-macro`

### DIFF
--- a/local-macros/src/lib.rs
+++ b/local-macros/src/lib.rs
@@ -1469,16 +1469,16 @@ pub fn enum_unwrapped(input: TokenStream) -> TokenStream {
 }
 
 fn get_node_enum_unwrapped_call(argument: &Expr, variant_name: &Ident) -> TokenStream2 {
-    let known_node_variant_names: HashMap<String, Vec<&'static str>> = HashMap::from_iter(
-        IntoIter::new([("Expression".to_string(), vec!["Expression"])]),
-    );
+    let known_node_variant_names: HashMap<String, Vec<&'static str>> =
+        HashMap::from_iter(IntoIter::new([
+            ("Expression".to_string(), vec![]),
+            ("VariableDeclarationList".to_string(), vec![]),
+        ]));
 
     let mut ancestors = vec!["Node"];
-    ancestors.extend_from_slice(
-        known_node_variant_names
-            .get(&variant_name.to_string())
-            .unwrap(),
-    );
+    let variant_name_string = variant_name.to_string();
+    ancestors.extend_from_slice(known_node_variant_names.get(&variant_name_string).unwrap());
+    ancestors.push(&variant_name_string);
     let ancestors = ancestors
         .into_iter()
         .map(|ancestor_str| Ident::new(ancestor_str, variant_name.span()));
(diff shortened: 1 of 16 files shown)


local-macros/src/lib.rs
```diff
@@ -1465,3 +1467,115 @@ pub fn enum_unwrapped(input: TokenStream) -> TokenStream {
...
+        let variant_name: Ident = input.parse()?;
+        Ok(NodeUnwrapped {
+            argument,
+            variant_name,
+        })
+    }
+}
+
+#[proc_macro]
+pub fn node_unwrapped(input: TokenStream) -> TokenStream {
```
helixbass, line 1514: Per comment on previous PR, I implemented `node_unwrapped!()`/`type_unwrapped!()` procedural macros on top of `enum_variant!()` that have knowledge of the `Node`/`Type` enum variants

But then I ended up feeling like the ergonomics of `.as_*()` were superior, and it seems nicer (both a bit faster to compile since it's only invoking one level of procedural macros instead of two and then the knowledge of the `Node`/`Type` variants lives outside of the procedural macro implementation) to implement those using `enum_unwrapped!()` rather than `node_unwrapped!()`/`type_unwrapped!()`

src/compiler/binder.rs
```diff
@@ -421,7 +420,8 @@ impl BinderType {
         name: __String,
     ) -> Rc<Symbol> {
         let symbol = self.create_symbol(symbol_flags, name).wrap();
-        enum_unwrapped!(&*self.file(), [Node, SourceFile])
+        self.file()
+            .as_source_file()
```
helixbass, line 424: The advantages of `.as_*()` are that (a) you don't need the `&*` boilerplate for `Rc`'s because they can deref to the `.as_*()` method calls on their inner `Node`/`Type` and (b) per one comment on previous PR, this puts things in more of a "pipeline" order

src/compiler/checker/lines_12000_16000.rs
```diff
@@ -386,7 +379,7 @@ impl TypeChecker {
                     } else {
                         TypeFlags::None
                     },
-                type_.as_union_or_intersection_type().types(),
+                type_.as_union_or_intersection_type_interface().types(),
```
helixbass, line 382: I ran into a naming conflict where `.as_union_or_intersection_type()` was the existing name for the helper for unwrapping to a `dyn UnionOrIntersectionTypeInterface`, so renamed it to `.as_union_or_intersection_type_interface()`

src/compiler/types.rs
```diff
@@ -1918,6 +1955,50 @@ impl Type {
             _ => panic!("Expected object flags type"),
         }
     }
+
+    pub fn as_base_interface_type(&self) -> &BaseInterfaceType {
```
helixbass, line 1959: I don't really mind having n of these, I'll just keeping adding them as-needed

src/compiler/binder.rs
peterstuart: Nice, this reads really nicely, even aside from the boilerplate reduction.

PR #44, opened 2022-01-13
Parse + bind type-alias declaration

helixbass:
In this PR:
- parse + bind a simple type alias declaration

To test:
```
type Num = number;

const a: Num = true;
```
then in the `post-binding:` debug-logging you should see `Num` show up in the `locals` symbol table of the `SourceFile` and the first statement should be a `TypeAliasDeclaration` with `name` `"Num"` and `type_` a `NumberKeyword` `KeywordTypeNode` (and those should have initialized `parent` bindings)

Based on `type-check-generic-interface-pr-feedback`

### DIFF
--- a/src/compiler/binder.rs
+++ b/src/compiler/binder.rs
@@ -345,6 +345,9 @@ impl BinderType {
             SyntaxKind::InterfaceDeclaration => {
                 return ContainerFlags::IsContainer | ContainerFlags::IsInterface;
             }
+            SyntaxKind::TypeAliasDeclaration => {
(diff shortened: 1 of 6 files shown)


src/compiler/binder.rs
```diff
@@ -382,6 +385,13 @@ impl BinderType {
                 symbol_flags,
                 symbol_excludes,
             )),
+            SyntaxKind::TypeAliasDeclaration => Some(self.declare_symbol(
```
helixbass, line 388: This isn't doing anything yet but I think this is where type parameters on generic type alias declarations will get put into its `locals` symbol table

PR #46, opened 2022-01-14
Parse block comment

helixbass:
In this PR:
- parse block comment
- include parsing diagnostics in emitted diagnostics

To test:
```
/* abc
 * def */
++a
```
then you should see a diagnostic debug-logged with message `Cannot find name 'a'.`
```
/* a
```
then you should see a diagnostic debug-logged with message `'*/' expected.`

Based on `type-check-type-alias`

### DIFF
--- a/src/compiler/types.rs
+++ b/src/compiler/types.rs
@@ -56,6 +56,7 @@ pub enum SyntaxKind {
     GreaterThanToken,
     MinusToken,
     AsteriskToken,
+    SlashToken,
     PlusPlusToken,
     MinusMinusToken,
     LessThanLessThanToken,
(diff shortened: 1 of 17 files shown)


src/compiler/types.rs
```diff
@@ -3157,7 +3239,7 @@ impl DiagnosticRelatedInformationInterface for DiagnosticRelatedInformation {
 #[derive(Clone, Debug)]
 pub struct BaseDiagnosticRelatedInformation {
     code: u32,
-    file: Option<Rc<SourceFile>>,
+    file: Option<Weak<SourceFile>>,
```
helixbass, line 3242: Ran into some circular debug-logging with this as a strong reference (specifically because `SourceFile` now has `parse_diagnostics`, which here circularly referred back to the `SourceFile`)

PR #48, opened 2022-01-15
Parse, bind + type-check var/ambient declarations

helixbass:
In this PR:
- parse, bind + type-check `var` declarations
- parse, bind + type-check "ambient" (`declare`) declarations

To test:
If you run `cargo run tmp.tsx` against a source file containing eg `var x: number = true;` then you should see a diagnostic debug-logged with message `Type 'boolean' is not assignable to type 'number'.`
```
declare var NaN: number;

const x: boolean = NaN;
```
then you should see a diagnostic debug-logged with message `Type 'number' is not assignable to type 'boolean'.`. In the `post-binding:` debug-logged `SourceFile` on the `VariableStatement`'s `BaseNode` you should see `modifiers` populated with a `DeclareKeyword` (and `parent` binding should be initialized)

Based on `parse-line-comment`


src/compiler/parser.rs
```diff
@@ -1707,20 +1762,65 @@ impl ParserType {
...
+    }
+
+    fn parse_declaration(&self) -> Statement {
+        let is_ambient = some(
+            self.look_ahead(|| {
+                self.parse_decorators();
+                self.parse_modifiers(None, None)
+            })
+            .as_ref()
+            .map(|node_array| {
```
helixbass, line 1768: This seemed necessary to convince it to treat an `&NodeArray` as an `&[Rc<Node>]` (which it derefs to now)

src/compiler/parser.rs
```diff
@@ -588,20 +536,30 @@ impl ParserType {
...
-    fn look_ahead<TReturn, TCallback: FnMut() -> Option<TReturn>>(
+    fn look_ahead<TReturn, TCallback: FnOnce() -> Option<TReturn>>(
         &self,
         callback: TCallback,
     ) -> Option<TReturn> {
         self.speculation_helper(callback, SpeculationKind::Lookahead)
     }
 
-    fn try_parse<TReturn, TCallback: FnMut() -> Option<TReturn>>(
+    fn look_ahead_bool<TCallback: FnOnce() -> bool>(&self, callback: TCallback) -> bool {
```
helixbass, line 546: The Typescript version of `lookAhead()`/`tryParse()` just looks at the falsiness of the callback return value, but in Rust sometimes that would correspond to an `Option` and sometimes to a `bool` so provide these `*_bool()` variants to make calling them more ergonomic in that case

PR #49, opened 2022-01-16
Parse + bind template literal

helixbass:
In this PR:
- parse + bind template literals

To test:
```
const a: boolean = `abc${"def"}`;
```
then you should see `post-binding:` debug-logging where the `initializer` of the `VariableDeclaration` is a `TemplateExpression` whose `head` is a `TemplateLiteralLikeNode` of `kind` `TemplateHead` and `text` `"abc"` and `template_spans` contains a `TemplateSpan` whose `expression` is a `StringLiteral` with `text` `"def"` and whose `literal` is a `TemplateLiteralLikeNode` of `kind` `TemplateTail` and `text` `""`, and these should have `parent` bindings initialized

Based on `parse-ambient-declaration`


src/compiler/parser.rs
```diff
@@ -501,9 +514,27 @@ impl ParserType {
         self.next_token_without_check()
     }
 
+    fn re_scan_template_token(&self, is_tagged_template: bool) -> SyntaxKind {
+        self.set_current_token(self.scanner().re_scan_template_token(
+            Some(&|message, length| self.scan_error(message, length)),
```
helixbass, line 519: Could expose an eg `.scanner_on_error()` method returning this wrapped closure (that gets injected as the scanner `on_error` callback where needed) to DRY this up if that didn't eg cause lifetime issues

src/compiler/parser.rs
```diff
@@ -613,6 +651,26 @@ impl ParserType {
...
     }
 
+    fn parse_expected_token(
+        &self,
+        t: SyntaxKind,
+        diagnostic_message: Option<&DiagnosticMessage>,
+        args: Option<Vec<String>>,
+    ) -> Node {
+        self.parse_optional_token(t).unwrap_or_else(|| {
+            enum_unwrapped!(
```
helixbass, line 661: Could add the same ergonomic eg `.as_template_literal_like_node()` methods to `MissingNode` as I've been doing for `Node`/`Type`

src/compiler/parser.rs
```diff
@@ -877,6 +966,43 @@ impl ParserType {
         }
     }
 
+    fn is_variable_declarator_list_terminator(&self) -> bool {
```
helixbass, line 969: These were in the wrong order

src/compiler/scanner.rs
```diff
@@ -69,6 +70,27 @@ pub fn token_to_string(t: SyntaxKind) -> Option<&'static String> {
     token_strings.get(&t)
 }
 
+fn is_line_break(ch: char) -> bool {
```
helixbass, line 73: Some of these were in the wrong place as `Scanner` methods instead of top-level function helpers

src/compiler/scanner.rs
```diff
@@ -416,6 +441,27 @@ impl Scanner {
...
     }
 
+    pub fn re_scan_template_token(
+        &self,
+        on_error: Option<ErrorCallback>,
+        is_tagged_template: bool,
+    ) -> SyntaxKind {
+        Debug_.assert(
+            self.token() == SyntaxKind::CloseBraceToken,
+            Some("'reScanTemplateToken' should only be called on a '}'"),
```
helixbass, line 451: Maybe this should use `'re_scan_template_token'` instead of `'reScanTemplateToken'`?

src/compiler/scanner.rs
```diff
@@ -714,3 +1120,11 @@ fn code_point_at(s: &str, i: usize) -> char {
 fn char_size(ch: char) -> usize {
     1
 }
+
+fn utf16_encode_as_string(code_point: u32) -> String {
+    char::from_u32(code_point).unwrap().to_string()
+}
+
+fn hex_digits_to_u32(str: &str) -> Result<u32, String> {
+    u32::from_str_radix(str, 16).map_err(|_| "Couldn't convert hex digits to u32".to_string())
```
helixbass, line 1129: This is mimicking the use of `parseInt(..., 16)` in the Typescript version

src/compiler/types.rs
```diff
@@ -1133,10 +1144,51 @@ pub trait LiteralLikeNodeInterface {
...
+    ) -> Self {
+        Self {
+            _literal_like_node: literal_like_node,
+            raw_text,
+            template_flags,
+        }
+    }
+}
+
+impl TemplateLiteralLikeNodeInterface for TemplateLiteralLikeNode {
```
helixbass, line 1177: This maybe doesn't need to be an interface (vs concrete stuff on `TemplateLiteralLikeNode` since it ended up being a concrete AST type and I don't think there will be other node types implementing this interface

PR #56, opened 2022-01-21
All top-level scanner helpers

helixbass:
In this PR:
- add all top-level helpers from scanner module
- add all `CharacterCodes`

To test:

Based on `all-node-tests`


src/compiler/scanner/lines_0_1000.rs
```diff
@@ -96,73 +940,644 @@ pub(super) fn is_code_point(code: u32) -> bool {
...
+lazy_static! {
+    pub(super) static ref shebang_trivia_regex: Regex = Regex::new(r"^#!.*").unwrap();
+}
+
+// pub(crate) fn is_shebang_trivia(
+//     text: &str,
+//     pos: usize, // this is a char index, not a string (byte) index, it's only ok because it's always 0 here
+// ) -> bool {
+//     Debug_.assert(pos == 0, None);
+//     shebang_trivia_regex.is_match(text)
```
helixbass, line 1185: Mimicking the Typescript version by using a regex for this shebang-processing was necessitating passing around the `&str` version of the source text (since that's what you need to feed to a regex), and then eg for `scan_shebang_trivia()` to return a # of _chars_ processed instead of a # of _bytes_ processed I needed to explicitly convert that. So since it's a simple regex to mimic by hand I just rewrote these to take a `&SourceTextAsChars` instead

src/compiler/types/lines_2000_4000.rs
```diff
@@ -534,6 +534,84 @@ impl TypeAliasDeclaration {
...
+pub fn text_str_num_chars(text: &str, start: usize, end: usize) -> usize {
+    text[start..end].chars().count()
+}
+
+pub trait SourceFileLike {
+    fn text(&self) -> &str;
+    fn text_as_chars(&self) -> &SourceTextAsChars;
+    fn maybe_line_map(&self) -> RefMut<Option<Vec<usize>>>;
+    fn line_map(&self) -> Ref<Vec<usize>>;
+    fn maybe_get_position_of_line_and_character(
```
helixbass, line 607: In the Typescript version `SourceFileLike` has an optional `getPositionOfLineAndCharacter()` method, I guess the most literal translation of that would be for `maybe_get_position_of_line_and_character` to be an `Option` of a closure type (that you can then call if it's present) but seems easier to have it always be implemented and just always return `None` if it's effectively not implemented

PR #58, opened 2022-01-21
All scanner methods

helixbass:
In this PR:
- all methods on `Scanner`

To test:

Based on `split-scanner-module`


src/compiler/scanner/lines_1000_2000.rs
```diff
@@ -23,44 +30,188 @@ impl Scanner {
...
+                on_error,
+                start,
+                Some(
+                    decimal_fragment.is_none()
+                        && self.token_flags().intersects(TokenFlags::Scientific),
+                ),
+            );
+            ScanNumberReturn {
+                type_: SyntaxKind::NumericLiteral,
+                value: result.parse::<f64>().unwrap().to_string(),
```
helixbass, line 150: The Typescript version uses `+` coercion to convert this string to a number, it looks like since Rust's `String.parse::<f64>()` can parse eg scientific notation it should be able to successfully convert from the number-string to a number

src/compiler/scanner/lines_1000_2000.rs
```diff
@@ -23,44 +30,188 @@ impl Scanner {
...
+            self.set_pos(identifier_start);
+        }
+    }
+
+    pub(super) fn scan_octal_digits(&self) -> u32 {
+        let start = self.pos();
+        while matches!(self.maybe_text_char_at_index(self.pos()), Some(ch) if is_octal_digit(ch)) {
+            self.increment_pos();
+        }
+        u32::from_str_radix(&self.text_substring(start, self.pos()), 8).unwrap()
```
helixbass, line 212: Here again the Typescript version uses `+` coercion, using `u32::from_str_radix()` should work for converting an octal/binary string into the corresponding number value

src/compiler/scanner/lines_1000_2000.rs
```diff
@@ -639,16 +1243,128 @@ impl Scanner {
                     return self.set_token(SyntaxKind::SemicolonToken);
                 }
                 CharacterCodes::less_than => {
+                    if is_conflict_marker_trivia(self.text(), self.pos()) {
+                        self.set_pos(scan_conflict_marker_trivia(
+                            self.text(),
+                            self.pos(),
+                            // Some(|diag, pos, len| self.error(on_error, diag, pos, len)),
```
helixbass, line 1250: For some reason it was complaining about lifetimes with this as an `Option` so I just made it non-optional and passed a dummy callback in places that would've passed `None`

PR #59, opened 2022-01-25
Some parenthesizer, utility functions

helixbass:
In this PR:
- some functions from factory parenthesizer module, utility modules

To test:

Based on `scanner-methods`


src/compiler/checker/lines_4000_8000.rs
```diff
@@ -318,15 +317,11 @@ pub(super) fn create_node_builder() -> NodeBuilder {
 }
 
 #[derive(Debug)]
-pub struct NodeBuilder {
-    synthetic_factory: BaseNodeFactorySynthetic,
```
helixbass, line 322: Updated the synthetic base node factory to be a `thread_local!` so updated to inject that value by calling `.with()` on the `thread_local!` at each point of usage

src/compiler/checker/lines_4000_8000.rs
```diff
@@ -379,107 +374,126 @@ impl NodeBuilder {
         context: &NodeBuilderContext,
     ) -> TypeNode {
         if type_.flags().intersects(TypeFlags::String) {
-            return Into::<KeywordTypeNode>::into(
-                factory
-                    .create_keyword_type_node(&self.synthetic_factory, SyntaxKind::StringKeyword),
-            )
+            return Into::<KeywordTypeNode>::into(synthetic_factory.with(|synthetic_factory_| {
+                factory.with(|factory_| {
```
helixbass, line 378: Also updated `factory` to be a `thread_local!`

src/compiler/command_line_parser.rs
```diff
@@ -1,6 +1,2980 @@
...
+            affects_semantic_diagnostics: None,
+            affects_emit: None,
+            affects_program_structure: None,
+            transpile_option_value: None,
+        })
+        .into();
+}
+
+thread_local! {
+    pub(crate) static jsx_option_map: HashMap<&'static str, JsxEmit> =
```
helixbass: This seemed to work fine, per a previous PR comment I should try and update the other static `HashMap`'s to use `&'static str`'s rather than allocating `String`'s for all of the entries

src/compiler/command_line_parser.rs
```diff
@@ -1,6 +1,2980 @@
...
+}
+
+thread_local! {
+    pub(crate) static lib_map: HashMap<&'static str, &'static str> = lib_entries
+        .with(|lib_entries_| HashMap::from_iter(lib_entries_.iter().map(|tuple| (tuple.0, tuple.1))));
+}
+
+thread_local! {
+    pub(crate) static common_options_with_build: Vec<Rc<CommandLineOption>> = vec![
+        CommandLineOptionOfBooleanType::new(CommandLineOptionBase {
```
helixbass: This seems like a builder pattern would be nice

src/compiler/types/lines_0_1000.rs
```diff
@@ -548,6 +586,10 @@ pub trait NodeInterface: ReadonlyTextRange {
     fn maybe_locals(&self) -> RefMut<Option<SymbolTable>>;
     fn locals(&self) -> RefMut<SymbolTable>;
     fn set_locals(&self, locals: Option<SymbolTable>);
+    fn maybe_js_doc(&self) -> Option<Vec<Rc<Node /*JSDoc*/>>>;
```
helixbass, line 589: These only apply to `Node` types that also `extend JSDocContainer` but seems to make sense to put them on `NodeInterface`/`BaseNode`

src/compiler/types/lines_2000_4000.rs
```diff
@@ -17,24 +18,34 @@ use local_macros::ast_type;
 pub struct BinaryExpression {
     pub _node: BaseNode,
     pub left: Rc<Node>,
-    pub operator_token: Box<Node>,
+    pub operator_token: Rc<Node>,
     pub right: Rc<Node>,
+    cached_literal_kind: Cell<Option<SyntaxKind>>,
```
helixbass, line 23: This only applies to `BinaryPlusExpression` but seemed to make sense to put it on `BinaryExpression`

src/compiler/types/lines_5000_5250.rs
```diff
@@ -34,6 +34,16 @@ impl __String {
...
         self.0.chars()
     }
+
+    // TODO: should implement via Deref instead?
+    pub fn as_str(&self) -> &str {
+        &self.0
+    }
+
+    // TODO: should implement via some trait?
+    pub fn eq_str(&self, str: &str) -> bool {
```
helixbass, line 44: I don't know if there's a standard trait for doing "equality comparison" against a different type?

src/compiler/types/lines_6000_8000.rs
```diff
@@ -231,4 +829,106 @@ pub enum EmitHint {
...
+
+        const Assertions = Self::TypeAssertions.bits | Self::NonNullAssertions.bits;
+        const All = Self::Parentheses.bits | Self::Assertions.bits | Self::PartiallyEmittedExpressions.bits;
+
+        const ExcludeJSDocTypeAssertion = 1 << 4;
+    }
+}
+
+pub trait ParenthesizerRules {
+    // fn get_parenthesize_left_side_of_binary_for_operator(&self, binary_operator: SyntaxKind) ->
```
helixbass, line 848: These methods returning closures seemed like they'd be tricky to implement on a trait like this, so for now planning on just "inlining" this optimization (I guess always just calling `parenthesize_left_side_of_binary()`?)

src/compiler/types/lines_8000_end.rs
```diff
@@ -19,12 +19,62 @@ pub struct PrinterOptions {}
...
+pub trait SymbolTracker {
+    fn track_symbol(
+        &mut self,
+        symbol: &Symbol,
+        enclosing_declaration: Option<Rc<Node>>,
+        meaning: SymbolFlags,
+    ) -> Option<bool> {
+        None
+    }
+    fn report_inaccessible_this_error(&mut self) {}
```
helixbass, line 50: For optional interface methods that return `void` or an `Option`, seems like giving a default trait method implementation is a nice way of not having to write out the "dummy" method implementations in individual implementors

PR #61, opened 2022-01-26
Some utilities public

helixbass:
In this PR:
- add some of `src/compiler/utilitiesPublic.ts` helpers

To test:


Cargo.toml
```diff
@@ -10,3 +10,5 @@ bitflags = "1.3.2"
 lazy_static = "1.4.0"
 local-macros = { path = "local-macros" }
 regex = "1.5.4"
+serde = "1.0.133"
```
helixbass, line 13: I think you can specify only needed parts of `serde` to make compilation faster?

local-macros/src/lib.rs
```diff
@@ -85,6 +85,14 @@ fn get_ast_struct_interface_impl(
                         self.#first_field_name.set_flags(flags)
                     }
 
+                    fn modifier_flags_cache(&self) -> crate::ModifierFlags {
```
helixbass, line 88: Pulled in some stuff from the `type-check-return-statement` branch

PR #62, opened 2022-01-28
All utilities public

helixbass:
In this PR:
- all helpers from `src/compiler/utilitiesPublic.ts`

To test:

Based on `some-utilities-public`


src/compiler/binder.rs
```diff
@@ -146,10 +145,10 @@ impl BinderType {
         self.Symbol()(flags, name).into()
     }
 
-    fn add_declaration_to_symbol<TNode: NodeInterface>(
```
helixbass, line 149: I got rid of this `<TNode: NodeInterface>` pattern everywhere since it tends to "infect" other code and I think it's cleaner to just always pass around `&Node` and then call `.as_*()` to get the "unwrapped" variants as needed (and even for functions that don't "infect" others, it's clunkier to pass an `Rc<Node>` (eg `&*node` just `&node`))

src/compiler/binder.rs
```diff
@@ -510,48 +509,39 @@ impl BinderType {
...
-            Node::PropertyAssignment(property_assignment) => self
-                .bind_property_or_method_or_accessor(
-                    property_assignment,
-                    SymbolFlags::Property,
-                    SymbolFlags::PropertyExcludes,
-                ),
-            Node::Statement(Statement::FunctionDeclaration(function_declaration)) => {
-                self.bind_function_declaration(function_declaration)
+        match node {
+            Node::TypeParameterDeclaration(_) => self.bind_type_parameter(node),
```
helixbass, line 513: Now that I'm always passing `&Node`, a lot of these can probably be `match`'s on `node.kind()` (closer to the Typescript version)

src/compiler/utilities_public.rs
```diff
@@ -621,93 +1057,450 @@ fn get_first_jsdoc_tag<TPredicate: FnMut(&Node /*JSDocTag*/) -> bool>(
...
+        Self::NodeArray(value)
+    }
+}
+
+pub fn get_text_of_jsdoc_comment<'a, TComment: Into<StrOrNodeArrayRef<'a>>>(
+    comment: Option<TComment>,
+) -> Option<Cow<'a, str>> {
+    match comment {
+        Some(comment) => match comment.into() {
+            StrOrNodeArrayRef::Str(comment) => Some(comment.into()),
```
helixbass, line 1100: You can use `.into()` to go from eg a `&str` to a `Cow`

PR #63, opened 2022-01-29
All parenthesizer rules

helixbass:
In this PR:
- all helpers from `src/compiler/factory/parenthesizerRules.ts`

To test:


src/compiler/factory/node_factory.rs
```diff
@@ -35,7 +39,7 @@ bitflags! {
     }
 }
 
-impl NodeFactory {
+impl<TBaseNodeFactory: 'static + BaseNodeFactory> NodeFactory<TBaseNodeFactory> {
```
helixbass, line 42: In order to have `ParenthesizerRules` be object-safe I couldn't have `<TBaseNodeFactory: BaseNodeFactory>` generics in the trait method signatures, so refactored so that `<TBaseNodeFactory>` is on the `NodeFactory`/`ParenthesizerRules` rather than on each individual method call

PR #65, opened 2022-02-01
All node converters

helixbass:
In this PR:
- all factory node converter helpers

To test:


src/compiler/debug.rs
```diff
@@ -1,12 +1,43 @@
...
 
-use crate::NodeArray;
+use crate::{AssertionLevel, Node, NodeArray};
+
+enum AssertionKeys {
+    AssertNode,
+}
+
+thread_local! {
+    static current_assertion_level: Cell<AssertionLevel> = Cell::new(AssertionLevel::None);
```
helixbass, line 14: Making this a struct member of `DebugType` would've made it non-thread-safe and I didn't want to make the ergonomics of using `Debug_` worse (eg by converting it to a `thread_local!`) so just made it a "standalone" global static

src/compiler/factory/node_converters.rs
```diff
@@ -0,0 +1,347 @@
...
+    cast, get_starts_on_new_line, is_array_binding_pattern, is_array_literal_expression,
+    is_binding_element, is_binding_pattern, is_block, is_expression, is_identifier,
+    is_object_binding_pattern, is_object_literal_element_like, is_object_literal_expression, map,
+    set_original_node, set_starts_on_new_line, set_text_range, BaseNodeFactory, Debug_,
+    FunctionLikeDeclarationInterface, HasInitializerInterface, HasTypeInterface,
+    HasTypeParametersInterface, NamedDeclarationInterface, Node, NodeConverters, NodeFactory,
+    NodeInterface, SignatureDeclarationInterface, SyntaxKind,
+};
+
+pub fn create_node_converters<TBaseNodeFactory: 'static + BaseNodeFactory>(
```
helixbass, line 14: Similar pattern as used for parenthesizer rules

src/compiler/factory/node_factory/lines_0_1000.rs
```diff
@@ -71,6 +95,26 @@ impl<TBaseNodeFactory: 'static + BaseNodeFactory> NodeFactory<TBaseNodeFactory>
...
+        Ref::map(self.converters.borrow(), |option| option.as_ref().unwrap())
+    }
+
+    pub fn create_assignment(
+        &self,
+        base_factory: &TBaseNodeFactory,
+        left: Rc<Node /*Expression*/>,
+        right: Rc<Node /*Expression*/>,
+    ) -> BinaryExpression {
+        self.create_binary_expression(base_factory, left, SyntaxKind::EqualsToken, right)
```
helixbass, line 115: There's some kind of currying optimization in the Typescript version but didn't attempt to mimic that here

PR #66, opened 2022-02-04
Some node factory helpers

helixbass:
In this PR:
- some node factory create helpers

To test:

Based on `node-converters`


src/compiler/utilities.rs
```diff
@@ -1959,16 +1992,23 @@ pub fn is_non_type_alias_template(tag: &Node /*JSDocTag*/) -> bool {
             }))
 }
 
-pub fn has_syntactic_modifier(node: &Node, flags: ModifierFlags) -> bool {
+pub fn has_syntactic_modifier<TNode: NodeInterface>(node: &TNode, flags: ModifierFlags) -> bool {
     get_selected_syntactic_modifier_flags(node, flags) != ModifierFlags::None
 }
 
-fn get_selected_syntactic_modifier_flags(node: &Node, flags: ModifierFlags) -> ModifierFlags {
+pub fn has_static_modifier<TNode: NodeInterface>(node: &TNode) -> bool {
```
helixbass, line 1999: One of these actually needed the `<TNode: NodeInterface>` pattern (vs passing `&Node`) because it was getting called prior to wrapping in `Node`

PR #69, opened 2022-02-06
All parser for each child

helixbass:
In this PR:
- all AST node types for parser `for_each_child()` helper

To test:

Based on `node-factory-create-helpers`


src/compiler/binder.rs
```diff
@@ -511,7 +510,7 @@ impl BinderType {
             Node::TypeParameterDeclaration(_) => self.bind_type_parameter(node),
             Node::ParameterDeclaration(_) => self.bind_parameter(node),
             Node::VariableDeclaration(_) => self.bind_variable_declaration_or_binding_element(node),
-            Node::TypeElement(TypeElement::PropertySignature(_)) => self.bind_property_worker(node),
```
helixbass, line 514: Got rid of `TypeElement` and `JSDocTag` enums

PR #77, opened 2022-02-20
JSDoc parser

helixbass:
In this PR:
- JSDoc parser

To test:

Based on `transformer-factory`


src/compiler/parser/lines_1000_2000.rs
```diff
@@ -61,15 +61,15 @@ impl ParserType {
             self.parse_statement()
         });
         Debug_.assert(self.token() == SyntaxKind::EndOfFileToken, None);
-        let end_of_file_token = self.add_jsdoc_comment(self.parse_token_node());
+        let end_of_file_token = self.add_jsdoc_comment(self.parse_token_node().into());
```
helixbass, line 64: `.add_jsdoc_comment()` ended up needing a wrapped `Rc<Node>` so that "infected" several parser method return types

src/rust_helpers/mod.rs
```diff
@@ -1,8 +1,22 @@
...
+use crate::SourceTextAsChars;
+
 pub mod number;
 pub mod weak_self;
 
 pub fn is_same_variant<TEnum>(value: &TEnum, other_value: &TEnum) -> bool {
     mem::discriminant(value) == mem::discriminant(other_value)
 }
+
+pub fn last_index_of(text: &SourceTextAsChars, ch: char, position: usize) -> isize {
```
helixbass, line 13: Mimic JS's `.lastIndexOf()` method

